PR #1, opened 2016-08-16
this one passed travis

magaliruffier:
see https://travis-ci.org/magaliruffier/ensembl-vep

### DIFF
--- a/travisci/get_dependencies.sh
+++ b/travisci/get_dependencies.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 echo 'Getting BioPerl'
-if [ ! -f release-1-6-924.zip ]; then
-  wget https://github.com/bioperl/bioperl-live/archive/release-1-6-924.zip
-  unzip -q release-1-6-924.zip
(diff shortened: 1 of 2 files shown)


willmclaren:
Yeah sorry to burst your bubble but they all pass now because I wrapped the use for Bio::DB::BigFile in an [eval](https://github.com/willmclaren/ensembl-vep/blob/master/modules/Bio/EnsEMBL/VEP/AnnotationSource/File.pm#L65-L70), and the tests [skip](https://github.com/willmclaren/ensembl-vep/blob/master/t/AnnotationSource_File_BigWig.t#L32) if it doesn't get included:

```
/home/travis/build/magaliruffier/ensembl-vep/t/AnnotationSource_File_BigWig.t ................... 
ok 1 - use Bio::EnsEMBL::VEP::AnnotationSource::File;
ok 2 # skip Bio::DB::BigFile module not available
ok 3 # skip Bio::DB::BigFile module not available
```
...etc...

magaliruffier:
that's why it suddenly started working yesterday when I pulled your latest changes, makes sense
will try digging further

PR #2, opened 2016-12-07
Use "defined" to test if there is a plugins list

Zhicheng-Liu:
The plugins key could be created somewhere else but without defined value. This
could cause issue if the plugins list is later de-referenced.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Runner.pm
+++ b/modules/Bio/EnsEMBL/VEP/Runner.pm
@@ -543,7 +543,7 @@ sub get_OutputFactory {
 sub get_all_Plugins {
   my $self = shift;
 
-  if(!exists($self->{plugins})) {
+  if(!defined($self->{plugins})) {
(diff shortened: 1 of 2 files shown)


willmclaren:
Can you create a test case that covers this situation and add it to t/Runner.t? It's hard from your comment to see exactly how this would happen?

willmclaren:
OK - but I still don't understand how you have got to this point that you need this fix? Can you please explain?

Zhicheng-Liu:
The problem appeared when I was configuring ensembl-vep for EG REST server. I did not trace back to see where the "plugins" are created but not defined. But I thought it would be good for VEP to handle this robustly.

willmclaren:
OK thanks that helps - the fix should go into REST, actually, I have found where the problem is.

I will close these PRs.

Zhicheng-Liu:
Well, I believe both client and API need fixing. But it is up to you.

PR #11, opened 2017-01-17
Update VCF.pm

at7:
Create individual vf only if genotype exists
- problem is caused by individuals with ./.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
@@ -356,6 +356,7 @@ sub create_individual_VariationFeatures {
 
     # get alleles present in this individual
     my $gt = $ind_gts->{$ind};
+    next if (!$gt);
     my @bits = split /\||\/|\\/, $gt;
     my $phased = ($gt =~ /\|/ ? 1 : 0);
 


at7:
I noticed another problem. If non of the samples have a genotype for the reported variant this code returns an empty array: (line~242)
if($self->{individual}) {
    return [
      map {@{$self->create_individual_VariationFeatures($_)}}
      @{$self->post_process_vfs([$vf])}
    ];
  }
And this makes the parser stop. I added a quick fix for this to my checkout: if map {@{$self->create_individual_VariationFeatures($_)}} @{$self->post_process_vfs([$vf])} is empty just return $self->post_process_vfs([$vf]).

willmclaren:
Fixed by https://github.com/Ensembl/ensembl-vep/commit/e1fe7731b44c1317be4f333ce198604b48c05364

at7:
Thank you. Another issue: In VCF::create_individual_VariationFeatures get_samples_genotypes doesn't correct alleles if they are indels. It returns TA/T instead of A/-. It gets then copied into the VF. Not sure if the alleles should be corrected in the BaseVCF4 parser in the first place.

willmclaren:
Fixed by https://github.com/Ensembl/ensembl-vep/commit/9d85aa8a4ae4b233777ff9e6572c87f640982512

PR #83, opened 2017-09-27
Use cached seq_region_start and seq_region_end

willmclaren:
Historically VEP has used $vf->{start} for coordinate lookup, and
mostly it works fine. Some of the TranscriptMapper methods
require $vf->seq_region_start(). Mostly this is OK as it just does
a slice-based lookup, but for circular slices e.g. in E.coli the
core API tries to do a DB lookup based on $vf->dbID, which
of course fails.

We get around it by caching seq_region_start on the VF itself.
This depends on an update the ensembl-variation
BaseVariationFeature module to implement that caching and avoid
defaulting to the parent method in Feature.

Have to be careful though, particularly when split_variants() and
minimise_alleles() create copies of VFs, hence the changes there.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/InputBuffer.pm
+++ b/modules/Bio/EnsEMBL/VEP/InputBuffer.pm
@@ -567,9 +567,9 @@ sub split_variants {
 
         # give it a new allele string and coords
         $new_vf->allele_string($ref.'/'.$alt);
-        $new_vf->{start} = $start;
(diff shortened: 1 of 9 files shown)


willmclaren:
See also https://github.com/Ensembl/ensembl-variation/pull/67

willmclaren:
Tests will fail until https://github.com/Ensembl/ensembl-variation/pull/67 gets merged

PR #89, opened 2017-10-23
Trigger dependent build on downstream ensembl-rest repo.

avullo:
The usual Travis CI build is altered so that another stage runs sequentially after the matrix build. You will now see two stages on the Travis interface: "Test" (matrix build) and "Trigger_dependent_buil". 
The new stage makes use of the Travis CI API to automatically trigger a build on the dependent repo. This does not occur in case another API triggered build is running or has run within the last hour.


avullo:
If this is ok, can I also alter core, variation, fungen and io in order to trigger a VEP build?

avullo:
For information, Build 659 stage 2 stopped because it's programmed not to run on PRs.

ens-lgil:
Looks ok to me. I believe we only need to check ensembl-rest as dependent repo

avullo:
Yes, I think so. I have analysed all dependencies between repos using the travis configuration files. vep is being checked out only by rest.

Thanks

avullo:
thanks @ens-lgil @willmclaren :+1:

Issue #97, opened 2017-11-07
Perl Warnings on Custom Annotation from gnomAD exomes


james-lawlor:
Hi,
When using VEP v. 90.5 to add custom annotations from the ensembl-provided gnomAD exome VCF, I'm running into a large number of perl warnings. I've got VEP set up to use nearly all the annotations listed in the gnomAD exome VCF header.

There do not seem to be any other negative impacts on the output, but with the number of warnings I thought it might be worth mentioning. I haven't yet been able to narrow down which annotation(s) may be responsible. (I haven't had the same problem from the genome VCF).

Here's there first 5000 warnings I get from annotating a typical VCF:

[warnings_5000.txt](https://github.com/Ensembl/ensembl-vep/files/1451249/warnings_5000.txt)

Toy example:
```
[jlawlor@login01 test_vep]$ vep -i tiny.vcf -o tiny_vep.vcf --config test_perl_errors.ini 
Use of uninitialized value $value in pattern match (m//) at /gpfs/gpfs2/software/vep-90.5/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm line 181, <__ANONIO__> line 54628.
Use of uninitialized value in join or string at /gpfs/gpfs2/software/vep-90.5/modules/Bio/EnsEMBL/VEP/Utils.pm line 183, <__ANONIO__> line 54628.
```
tiny.vcf:
```
1       13656   .       CAG     C       54267.10        VQSRTrancheINDEL99.00to99.90
```

test_perl_errors.ini:
```
[...]

PR #98, opened 2017-11-09
addressing https://github.com/Ensembl/ensembl-vep/issues/97

willmclaren:


willmclaren:
Also addresses https://github.com/Ensembl/ensembl-vep/issues/99

Issue #108, opened 2017-11-25
filter_vep ignores VEP annotations when SnpEff annotations are available as well


Oodnadatta:
Hello VEP team and thank you for this tool!

I have a vcf with SnpEff and VEP annotations.

    ##INFO=<ID=CSQ,Number=.,Type=String,Description="Consequence annotations from Ensembl VEP. Format: Allele|Consequence|IMPACT|SYMBOL|Gene|Feature_type|Feature|BIOTYPE|EXON|INTRON|HGVSc|HGVSp|cDNA_position|CDS_position|Protein_position|Amino_acids|Codons|Existing_variation|DISTANCE|STRAND|FLAGS|VARIANT_CLASS|SYMBOL_SOURCE|HGNC_ID|GIVEN_REF|USED_REF|SIFT|PolyPhen|AF|AFR_AF|AMR_AF|EAS_AF|EUR_AF|SAS_AF|AA_AF|EA_AF|gnomAD_AF|gnomAD_AFR_AF|gnomAD_AMR_AF|gnomAD_ASJ_AF|gnomAD_EAS_AF|gnomAD_FIN_AF|gnomAD_NFE_AF|gnomAD_OTH_AF|gnomAD_SAS_AF|CLIN_SIG|SOMATIC|PHENO">

[...]

PR #111, opened 2017-11-28
Refine filter_vep handling of VCF

willmclaren:
filter_vep will now by default parse VEP annotations encoded in the CSQ
INFO field of a VCF; this can be configured (e.g. to use ANN) with
--vcf_info_field.

Addresses https://github.com/Ensembl/ensembl-vep/issues/108

### DIFF
--- a/filter_vep
+++ b/filter_vep
@@ -62,6 +62,7 @@ sub configure {
     start => 1,
     limit => 1e12,
     output_file => 'stdout',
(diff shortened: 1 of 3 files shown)


willmclaren:
Also added another commit now addressing missing cell types in output when using `--cell_type`

willmclaren:
Ah yes, thanks I added it to the web docs but not to the help message.

PR #121, opened 2018-01-09
Revert "fix Uploaded_variation"

at7:


monsanto-pinheiro:
Why did you revert "fix Uploaded_variation"?

at7:
The pull request is against release/91. But we need to add the code to master. The change will be part of release/92. Could you please commit against master?

PR #122, opened 2018-01-10
update to handle RefSeq MT names

sarahhunt:
Also required a number of unrelated test updates to deal with the additional chromosome in the cache.


modules/Bio/EnsEMBL/VEP/AnnotationType/Transcript.pm
```diff
@@ -210,14 +210,14 @@ sub filter_transcript {
       # we only want RefSeq transcripts e.g. NM_12930
```
helensch: Can a comment be added about MT transcripts having 4 digit stable ids and not of format NM_

sarahhunt: Thanks @helensch - good point.

PR #124, opened 2018-01-16
Added ambiguity flag

aparton:
Added ambiguity flag into CL and REST VEP. Changes to ensembl-vep, ensembl-rest and public_plugins have all been made at the same time.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -99,7 +99,6 @@ our %DEFAULTS = (
   terminal_width    => 48,
   vcf_info_field    => 'CSQ',
   ucsc_data_root    => 'http://hgdownload.cse.ucsc.edu/goldenpath/',
-  ambiguity         => 0,
   
   # frequency filtering
   freq_freq         => 0.01,
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -99,6 +99,7 @@ our %DEFAULTS = (
   terminal_width    => 48,
   vcf_info_field    => 'CSQ',
   ucsc_data_root    => 'http://hgdownload.cse.ucsc.edu/goldenpath/',
+  ambiguity         => 0,
   
   # frequency filtering
   freq_freq         => 0.01,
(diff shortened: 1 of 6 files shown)


modules/Bio/EnsEMBL/VEP/Config.pm
```diff
@@ -99,6 +99,7 @@ our %DEFAULTS = (
   terminal_width    => 48,
   vcf_info_field    => 'CSQ',
   ucsc_data_root    => 'http://hgdownload.cse.ucsc.edu/goldenpath/',
+  ambiguity         => 0,
```
willmclaren: This isn't necessary; a default value of `undef` will be assumed for any parameter that doesn't get a value in DEFAULTS, and this fine for any flag that should be off by default.

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -1470,6 +1471,10 @@ sub TranscriptVariationAllele_to_output_hash {
     }
   }
 
+  if($self->param('ambiguity') && defined($vfoa->variation_feature->ambig_code())) {
```
willmclaren: You've added `ambiguity` to the param shortcuts for this object, so you can access its value with `$self->{ambiguity}` rather than the somewhat slower `$self->param('ambiguity')`.

It would also be better to write this such that you're fetching the value of `$vfoa->variation_feature->ambig_code()` only once by assigning it to a local variable then checking it.

Because this value is not transcript-specific, its retrieval doesn't belong here either; it should be placed in `VariationFeature_to_output_hash`  instead so that it is only called once per VariationFeature rather than once every TranscriptVariation as here.

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -807,6 +808,12 @@ sub VariationFeature_to_output_hash {
 
   # minimised?
   $hash->{MINIMISED} = 1 if $vf->{minimised};
+  
+  my $ambiguity_code = $vf->ambig_code();
+  
+  if($self->{ambiguity} && defined($ambiguity_code)) {
```
at7: ambig_code is not implemented for structural variants. You could test if the variant is a structural variant with $vf->{overlapping_svs}. if not you can get the ambig code

willmclaren: This is true, but checking `$vf->{overlapping_svs}` won't work; this key contains overlapping structural variants added when `--check_svs` is enabled.

You should check the class of $vf using `ref()` or `$vf->isa()` instead.

PR #126, opened 2018-01-18
Handled length exception more cleanly

aparton:
Prevents whole exception messages being printed to screen with common errors

Other messages can be added to the known_messages array to remove those exception messages also.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
@@ -193,12 +193,9 @@ sub create_VariationFeatures {
   }
 
   unless(@$vfs || scalar(@errors) == 0) {
-    #my @known_messages = ('MSG: Region requested must be smaller than 5kb');
-    $DB::single = 1;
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
```diff
@@ -193,12 +193,9 @@ sub create_VariationFeatures {
   }
 
   unless(@$vfs || scalar(@errors) == 0) {
-    #my @known_messages = ('MSG: Region requested must be smaller than 5kb');
-    $DB::single = 1;
     my %known_messages_hash = ('MSG: Region requested must be smaller than 5kb' => 0);
```
ens-lgil: Shouldn't it be with the value 1 ?

aparton: As far as I know it shouldn't matter - I'm not using the value assigned here, I'm simply using the hash as it's faster to check for the existence of a key rather than grep-ing through an array, so that 0 is just a dummy value. I might be missing something however.

ens-lgil: I am fine if it works on your tests!

PR #131, opened 2018-01-23
Fix missing ';' character when the INFO column of the input VCF conta…

ens-lgil:
…ins other fields after an existing CSQ field


ens-lgil:
This will need to be cherry-picked into the branch release/92

PR #138, opened 2018-01-31
update to handle RefSeq MT names

sarahhunt:
Copying changes which are already in master & 92 to the currrent release


ens-lgil:
Failing travis is due to a change in the ensembl-funcgen (master branch). This is fine for the fix on release/91

PR #145, opened 2018-02-20
remove temp code to calculate missing MAF

sarahhunt:


ens-lgil:
Change merged into release/92 and master

PR #146, opened 2018-02-21
Allows --AUTO flag to be fully automatic with --DESTDIR flag and on OSX

aparton:
Concerning https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-998

Additionally, a similar issue would occur when installing on OSX, so that case is being handled as well


sarahhunt:
@ens-lgil - yes!

ens-lgil:
Merged to release/92 and master

PR #158, opened 2018-03-15
Update VEP API and tests regarding the Ensembl Core API and schema ch…

ens-lgil:
…anges on master


ens-lgil:
Tests failing with Perl 5.8. The next PR will remove the test on Perl 5.8 (not supported anymore by the others Ensembl APIs).

ens-lgil:
Actually we might not need all of these changes

PR #163, opened 2018-03-20
Fix throw argument for existing output file

helensch:
Fix warning that argument to throw was not numeric
Bio::EnsEMBL::Utils::Exception::throw          
  Arg [1]    : string $msg
  Arg [2]    : (optional) int $level


helensch:
Move pull request to postreleasefix/92

PR #171, opened 2018-03-23
Patch the Regulation test database

ens-lgil:
This is only a SQL schema patch.
Further work will be needed:

- Add/move test data in the new SQL tables.
- Update the ensembl-vep API and tests


ens-lgil:
Cancelled for this release

PR #174, opened 2018-03-27
Fix issue caused by a code update about Tabix in https://github.com/E…

ens-lgil:
…nsembl/Bio-DB-HTS


at7:
I don't understand why you need the if clause. I can see that seek returns 0 if there are no more lines to iterate. Can you please add a test to illustrate what goes wrong without using if. If possible? Or at least add a comment what goes wrong without if?

ens-lgil:
Merged to `release/92` and `master`

PR #187, opened 2018-04-11
Improve Dockerfile: configure locales and a proper /opt installation dir

juanesarango:
We use VEP in an HPC cluster that doesn't support docker. We use [singularity](http://singularity.lbl.gov/) instead, which supports running docker images. Currently, we have some issues with the conversion from the docker images, because of the following definitions in the Dockerfile:
1) The locales are no set properly.
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LC_COLLATE = "C",
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```

2) In singularity $HOME is inherited from the host, so we need some cumbersome settings to run the containers in other folder. And the Dockerfile is installing everything in `/home`.

In this PR, I made the following improvements for a _**better docker configuration**_, that hope will benefit people that uses singularity, or docker with different $HOME settings: 

**installing and setting up the locales and changing the installation directory from $HOME (`/home`) to $OPT (`/opt`)**

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,32 +3,50 @@ FROM ubuntu
 
 # update aptitude and install some required packages
 # a lot of them are required for Bio::DB::BigFile
-RUN apt-get update && apt-get -y install apache2 build-essential cpanminus curl git libmysqlclient-dev libpng12-dev libssl-dev manpages mysql-client openssl perl perl-base unzip vim wget
+RUN apt-get update && apt-get -y install \
+    apache2 \
+    build-essential \
+    cpanminus \
+    curl \
+    git \
+    libmysqlclient-dev \
+    libpng12-dev \
+    libssl-dev \
+    locales \
+    manpages \
+    mysql-client \
+    openssl \
+    perl \
+    perl-base \
+    unzip \
+    vim \
+    wget
+
 # install ensembl dependencies
 RUN cpanm DBI DBD::mysql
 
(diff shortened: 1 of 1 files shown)


ens-lgil:
Dear @juanesarango,

Many thanks for this improvement.
I hope that this fix will be helpful for the other VEP users.

Best regards,
Laurent

juanesarango:
Hi @ens-lgil,
Thanks for accept my PR. 

I noticed it's necessary to update the [VEP download](https://useast.ensembl.org/info/docs/tools/vep/script/vep_download.html#docker)  documentation with docker as well. 

There are **3 instructions** where they suggest to use this flag: 
`-v $HOME/vep_data:/home/vep/.vep`

`/home/vep` is suggested as it was the previous $HOME. Now it should be `/opt/vep` as is the new $HOME inside the container, and VEP looks for $HOME/.vep as the default directory for cache and plugins. 

So, is **just changing** this in the docs: 
- `-v $HOME/vep_data:/home/vep/.vep` by `-v $HOME/vep_data:/opt/vep/.vep`

ens-lgil:
Hi @juanesarango,

Thank you for the reminder!
The web documentation has been fixed and should be live today.

Best regards,
Laurent

PR #188, opened 2018-04-12
Prf update json out

at7:
Update vep REST output keys from maf to af. We already changed the naming for the VEP script and tool output.


ens-lgil:
Changes merged into `master` and `release/93`

at7:
Thanks for the warning. The initial [matching](https://github.com/Ensembl/ensembl-vep/blob/release/92/modules/Bio/EnsEMBL/VEP/AnnotationType/Variation.pm#L88) gave me some false security. Ideally, I can use OutputFactory::add_colocated_frequency_data for adding frequency data to the JSON. The plan is to keep the structure of the VEP REST output format but make sure to correctly return the allele frequency for the input allele.

PR #202, opened 2018-05-01
Af match json prf93

at7:
Report allele frequency for variant allele in REST JSON output. This is for release/93 and won't contain max population allele frequencies.


modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm
```diff
@@ -355,57 +360,48 @@ sub add_VariationFeatureOverlapAllele_info {
...
 sub add_colocated_variant_info_JSON {
   my $self = shift;
-  my $vf = shift;
   my $hash = shift;
+  my $frequency_hash = shift;
+  my $ex_orig = shift;
+  
+  # work on a copy as we're going to modify/delete things
+  my $ex;
+  %$ex = %$ex_orig;
```
ens-lgil: Could you just use `my %ex = %$ex_origin;` (and use $ex{...} instead of $ex->{...} in the rest of the method) ?

ens-lgil: Sorry I didn't see that it was already there in the code before your changes

modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm
```diff
@@ -235,6 +234,12 @@ sub get_all_output_hashes_by_InputBuffer {
       delete $hash->{$key};
     }
 
+    my $variant_allele = $hash->{transcript_consequences}[0]->{variant_allele};
```
willmclaren: This won't work in all cases for two reasons:

- `$hash->{transcript_consequences}` isn't guaranteed to be populated
- variants can have more than one ALT, VEP should report frequencies for all ALTs

at7: I keep ignoring the fact that there can be more than one ALT allele. Core already excepted the change from *_maf to *_af! Which means users are most likely to loose information if they retrieve data by *_maf and haven't checked REST's change log. We might as well be even more radical with the format change for next release.  Do you have any comments for the following format suggestion?:
colocated_variants: [
   frequencies: {
      A: {
        amr: 0.01,
        gnomad_eas: 0.001,
...
      },
     G:  {
       amr: 0.3,
       gnomad_eas: 0.33,
...
     }
   },
  phenotype_or_disease: 1,
  seq_region_name: 2,
...
]

willmclaren: I like that, very sensibly laid out, avoids duplication of the non-allele-specific info.

PR #203, opened 2018-05-03
add variant name to warning message

ima23:


modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -593,12 +593,16 @@ sub validate_vf {
 
   # insertion should have start = end + 1
   if($vf->{allele_string} =~ /^\-\// && $vf->{start} != $vf->{end} + 1) {
+    my $variant_name = "";
+    if (defined $vf->name){
+      $variant_name = "for variant (".$vf->name.") ";
+    }
```
ens-lgil: Trivial but we could use the more condensed if/else syntax:
`my $variant_name = (defined $vf->name) ? "for variant (".$vf->name.") " : "";`

modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -593,12 +593,16 @@ sub validate_vf {
 
   # insertion should have start = end + 1
   if($vf->{allele_string} =~ /^\-\// && $vf->{start} != $vf->{end} + 1) {
+    my $variant_name = "";
+    if (defined $vf->name){
+      $variant_name = "for variant (".$vf->name.") ";
```
ens-lgil: Maybe we could remove the parenthesis, and use single quotes instead

modules/Bio/EnsEMBL/VEP/Parser.pm
ima23: Good suggestion, thanks! I've update it.

modules/Bio/EnsEMBL/VEP/Parser.pm
ima23: I went with the use of parentheses as they are used in the rest of the message seems to use parentheses to show the specific data. I can change this for variant name only if that is the standard. Is the use of single quotes in the variant name case only the standard way to display it?

PR #213, opened 2018-05-22
add OverlapBP and OverlapPC constants

ima23:
- add OverlapBP, OverlapPC to the constants


ens-lgil:
So the update is not scheduled for e93 ?

ens-lgil:
Commit cherry picked to `release/93`

PR #216, opened 2018-05-25
Add  Ensembl::XS to  docker

juanesarango:
According to one of the recommendations (# 9) to run [VEP faster](https://ensembl.org/info/docs/tools/vep/script/vep_other.html#faster), is to use perl package Ensembl::XS. I'm not sure why it doesn't come installed by default. But I thought it's a good idea to install it in the docker image as it works without issue, and it optimizes the runtime.

Please let me know what you think.


ens-lgil:
Dear @juanesarango,

Many thanks for your suggestion!
We try to improve/develop the Docker image for VEP (little by little) and the use of `ensembl-xs` is one of them.
I slightly alter your suggested change (PR #219 ) by using a defined version of the `ensembl-xs` repository (actually the latest) as its `master` branch corresponds to a development branch.

Best regards,
Laurent

PR #220, opened 2018-05-30
Clear registry before running VEP

at7:
The QC step was failing because during a VEP run errors like _No sequence found for seq_region 13699:62522_ occurred. At the same time we would get warnings: _WARN: Species (homo_sapiens) and group (core) same for two separate databases_ There could have been a conflict between different core databases (core and otherfeature maybe). Calling clear on the registry is solving the problem. Although, I don't know exactly why.


at7:
Thanks. I'll merge it to master and release/93.

PR #222, opened 2018-05-31
remove variants which no longer have 1KG frequencies

at7:
Both variants no longer have 1KG frequencies in the updated files provide by EVA. This is expected and has been discussed with Susan. We are still waiting for recalled 1KG data on GRCh38 :-)

The QC Pipeline should now pass the QC steps and if not it really means that something went wrong.


at7:
Also merged to master and release/93

PR #229, opened 2018-07-02
Master xref

aparton:
Updates to VEP Dump pipeline for new Xref changes

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
@@ -564,6 +564,7 @@ sub prefetch_gene_ids {
   }
 
   if( $self->{core_type} eq 'otherfeatures'){
+    #Pulls correct gene symbol for RefSeq using xrefs
     my @entries = grep {$_->{dbname} eq 'EntrezGene'} @{$tr->get_Gene()->get_all_DBEntries};
     if(scalar @entries eq 1)
     {
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
@@ -245,6 +245,7 @@ sub get_features_by_regions_uncached {
 
       foreach my $tr(@{$gene->get_all_Transcripts}) {
         next unless $self->filter_transcript($tr);
+        next if $tr->analysis() && $tr->analysis()->logic_name() eq 'estgene';
         # in human and mouse otherfeatures DB, there may be duplicate genes
         # skip those from analysis refseq_human_import and refseq_mouse_import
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/VEP/AnnotationType/Transcript.pm
```diff
@@ -224,7 +224,7 @@ sub filter_transcript {
   ) {
     return 0;
   }
-
+  return 0 if $t->analysis() && $t->analysis()->logic_name() eq 'estgene';
```
sarahhunt: I expected to find this around Line 250 in modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm. The EST genes are not useful, so we don't need them in cache at all.

modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
```diff
@@ -646,6 +659,12 @@ sub prefetch_translation_ids {
   # Ensembl protein ID
   if($self->{protein}) {
     $tr->{_protein} = $tl->stable_id;
+    my @entries = grep {$_->{dbname} eq 'GenBank'} @{$tl->get_all_DBEntries};
```
sarahhunt: Generally, more comments would be good. This definitely needs some comments as it looks as if odd things are happening to retrieve the Ensembl protein id, given the current comments.

modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
```diff
@@ -245,16 +244,30 @@ sub get_features_by_regions_uncached {
...
-        $tr->{_gene_stable_id} = $gene_stable_id;
+        if(defined($tr->display_xref) and $self->{core_type} eq 'otherfeatures'){
+	         $tr->{stable_id} = $tr->display_xref->{display_id};
+	      }
+        
+	      $tr->{_gene_stable_id} = $gene_stable_id;
         $tr->{_gene} = $gene;
         $self->prefetch_gene_ids($tr);
-
+        if( $self->{core_type} eq 'otherfeatures'){
```
sarahhunt: Can this section go in prefetch_gene_ids?

aparton:
Hiya,

Responding to each of these:

1) EST Genes - Is it not doing pretty much that? It was in get_features_by_regions_uncached initially, but line 247 of AnnotationSource/Database/Transcript.pm calls this filter_transcript method, so I just moved it inside. I might be misunderstanding how this code works though, is this method not a very early check to prevent adding things to the cache that we don’t want? I’ve left the check in filter_transcript for now, but I can move it back out into get_features_by_regions_uncached if you think it’s cleaner.

2) Indentation - Fair point. I’m blaming Atom again. It’s got a bucketload of cool features, but when it messes around with whitespace it drives me mad. Still haven’t found the setting(s) to stop that.

3) Swapping the display_xref and other features check - Good idea. Done.

4) Comments - added.

5) Adding the check to prefetch_gene_ids - I think yes, but I’ve just added it and now the output is different and the gene_ids aren’t all there. Not sure why that is, perhaps the $tr objects are slightly different? I’ll take a quick look

Thanks for these,
Andy

PR #230, opened 2018-07-02
Fix 'pick' options issues for structural variants

ens-lgil:


ens-lgil:
@willmclaren possibly, but the issue is that the code was looking for the method `variation_feature_seq` which doesn't exist in `StructuralVariationOverlapAllele.pm`.
I believe I need to define an "allele" for the SVs (e.g. https://github.com/Ensembl/ensembl-vep/blob/release/92/modules/Bio/EnsEMBL/VEP/OutputFactory.pm#L545).

willmclaren:
Sure. My opinion would be you have to think what the user would anticipate would happen in this scenario. Since SVs can only have one alt allele (I think???), then any use of e.g. `--pick_allele` should revert to the basic `--pick` behaviour for SVs (and SVs only, obviously).

It might be cleaner to write a filter method just for SVs that merges the behaviour of `--pick_allele` + `--pick` etc.

Happy to do it when I get a free minute if you're stuck.

ens-lgil:
Hi @willmclaren, yes you are right.
As we were supposed to branch yesterday (to 94), I was focusing on fixing the "syntax" error but I was planning to implement what you proposed in the next days.
Last night I was thinking about replicating how the `--pick` flag works for SVs, but maybe a filter would work better. I believe the issue is not only with `--pick_allele` but with most of the `--pick_*` and `--flag_pick_*` flags.

PR #238, opened 2018-07-12
Fix an issue with the flag '--shift_hgvs'

ens-lgil:


ens-lgil:
Merged to `release/94` and `master`

PR #240, opened 2018-07-12
Fix an issue with the flag '--shift_hgvs'

ens-lgil:


aparton:
Happy enough with these 3, although I do get the warning message:

"my" variable $vfoa masks earlier declaration in same scope at t/OutputFactory.t line 1919.

When running 'perl t/OutputFactory.t

ens-lgil:
Ok I will fix it

PR #253, opened 2018-08-01
Fix missing population frequency when the flag '--filter_common' is used

ens-lgil:


at7:
Can you please remind me why you are deleting the option?

ens-lgil:
This option removed the population frequency (1KG, gnomAD,...) when we used the `--filter_common` flag.
This might be due to an older setting in the former VEP code

ens-lgil:
@ima23 did you merged it to `release/94` as well ?

ima23:
I've merged it on release/94 too. Thanks for checking!

PR #255, opened 2018-08-01
Replace the default value for the FREQS filter to 'NA' when no data i…

ens-lgil:
…s available from the cache


ens-lgil:
Merged into `release/94` and `master`

PR #257, opened 2018-08-09
Cache creation: do not over-write tabixed file with non-tabixed file

sarahhunt:
Fix to ensure tabix'ed tarball destined for web/REST is not over-written by the non-tabix'ed version.

Also remove REST directory as it is no longer required.


at7:
merged to release/94 and master

PR #261, opened 2018-08-17
expand halo json output docs

ima23:


ima23:
added to release/94 and master

PR #262, opened 2018-08-21
warn which format has been detected when using verbose

sarahhunt:
Help with debuging unusual input data


ens-lgil:
Merged into `release/94` and `master`

PR #263, opened 2018-08-21
Vep dump changes 94

at7:
There is no fix ready for this bug: https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1172
Run without the error causing tests for release/94
Don't dump MotifFeature for release/94


ens-lgil:
If the fix is only meant for `release/94`, the PR should be on `release/94` and not on `postreleasefix/94` as we use the branches `postreleasefix/XX` to populate changes/fixes on both `master` and `release/XX`.

at7:
Thanks for the reminder. This goes to master too (for now).

ens-lgil:
Ok, it makes sense to be on the safe side, just in case the fix won'tbe available for e95 ...

PR #264, opened 2018-08-22
download correct htslib version

at7:
download htslib from release page


willmclaren:
I'm not sure but I think this will probably fail on a good number of user systems that are unable to unpack `.bz2` archives.

ens-lgil:
Do we still need it on `release/93` ?

at7:
It was actually not fixing the problem of installing htslib and biodbhts with vep install script.

PR #270, opened 2018-08-29
Master

aparton:
Updating assembly names when guessing in install file


ens-lgil:
I am going to close this PR and create a new one on `postreleasefix/94` with only 1 commit instead of 4.

PR #271, opened 2018-08-31
Fix alias discrepancy between the VEP web doc and the scripts

ens-lgil:


sarahhunt:
added to release/94 and master

PR #272, opened 2018-09-06
Htslib biodbhts update 94

at7:
Moving to most recent versions of htslib and Bio-DB-HTS. Same version update as we are going to have on the farm.


ens-lgil:
Is it working ?

PR #273, opened 2018-09-10
Allele string fix

sarahhunt:
Clean up allele strings for e!94 and add error catching around allele matching code to report any problem variants.


sarahhunt:
merged to master and release/94

PR #276, opened 2018-09-24
Ignore HGVS input that include square brackets

aparton:
PR to filter out user-given HGVS that includes square brackets - causing issues on web


modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
```diff
@@ -154,6 +154,11 @@ sub create_VariationFeatures {
 
   # remove whitespace
   $hgvs =~ s/\s+//g;
+  
+  if ((index($hgvs, '[') != -1) || (index($hgvs, ']') != -1)) {
```
ens-lgil: Maybe a regex would be more generic/compact in order to add easily more conditions in the future?
e.g. `if ($hgvs =~ /(\[|\])/) {`

PR #283, opened 2018-09-27
Coding only changes to prevent missing data when whole input buffer s…

aparton:
…lice is empty


aparton:
Anja fixed this issue while dealing with another bug. Closing the PR.

PR #288, opened 2018-10-01
Remove archives after uncompression to save some disk space (mainly f…

ens-lgil:
…or the Docker build)


ens-lgil:

PR #290, opened 2018-10-03
Rewrite of the Dockerfile:

ens-lgil:
- Use less layers by grouping the commands
- Use multi-stage build to reduce layers and final image size
- Try to cleanup the librairies and directories in order to get a
  smaller final image (about 30% smaller)


ens-lgil:

PR #293, opened 2018-10-08
Add a new option to return the reference allele in the TAB and VEP ou…

ens-lgil:
…tput formats


ens-lgil:
No problem @at7! I
Thank you for the comment: I wasn't 100% of the option and field names.

PR #300, opened 2018-10-11
Add an additional chromosome name check for the custom files with chr…

ens-lgil:
…omosome names like 'chr1'


ens-lgil:

PR #305, opened 2018-10-12
Remove warning message when the 'record ID' doesn't exist in the 'if'…

ens-lgil:
… statement


ens-lgil:

PR #308, opened 2018-10-17
Add a new option to return the reference allele in the TAB and VEP output formats

ens-lgil:
With the use of REF_ALLELE instead


ens-lgil:
@aparton, the reference allele is present by default in the VCF format, that's why I didn't add it for this format

PR #309, opened 2018-10-18
Make Dockefile changes: use Ubuntu in first stage, move the compilati…

ens-lgil:
…on of some libraries in the first stage, make the Docker image smaller by removing the 'kent-335_base' libraries not used in VEP.
It also add the LD C scripts (required to run the LD plugin)


ens-lgil:

PR #312, opened 2018-10-30
Feature/xrefsymbol

aparton:


ens-lgil:
@aparton can you add a commit to increment the VEP subversion in the `Constants.pm`, please ?
https://github.com/Ensembl/ensembl-vep/blob/release/94/modules/Bio/EnsEMBL/VEP/Constants.pm#L56

PR #324, opened 2018-11-14
Add missing general license for the repo

ens-lgil:


ens-lgil:
It's the standard LICENSE file of the project. I simply c/p from the `ensembl-variation` repo

PR #328, opened 2018-11-20
Add motif feature 95

at7:
We are adding motif features back to VEP caches. We only add experimentally validated motif features.


modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/Dumper/Regulation.pm
```diff
@@ -36,7 +36,7 @@ use Scalar::Util qw(weaken);
 use Bio::EnsEMBL::VEP::Config;
 use Bio::EnsEMBL::VEP::AnnotationSource::Database::RegFeat;
 use Bio::EnsEMBL::VEP::AnnotationSource::Cache::RegFeat;
-
+use Data::Dumper;
```
helensch: Is this needed?

PR #333, opened 2018-11-28
Fix VCF output missing VCF format header if the header was missing in…

ens-lgil:
… the input file


ima23:
merged to `master` and `release/95`

PR #338, opened 2018-12-04
From release 95 onward, all the ensembl databases have a division

thomasmaurel:
@at7: This is regarding the issues with dump path for vertebrates in release 95. The code is not broken but because we have updated the division for vertebrates, we should run the pipeline with the division flag in the future. I've updated the code to make things a bit cleaner and will update the documentation once this PR is merged.


at7:
Will be merged with #349. Which makes it easier for us to move the change to master and release/95 branches. Thank you for the PR.

PR #339, opened 2018-12-04
Fix issue to install the cpan module Bio::DB::BigFile

ens-lgil:


ens-lgil:
Merged to `master` and `release/95`

PR #341, opened 2018-12-05
Add the BIOTYPE field when the flag '--regulatory' is used

ens-lgil:


ens-lgil:
Merge into `master` and `release/95`

PR #351, opened 2018-12-13
Add docker tag

ens-lgil:


ens-lgil:
Here is the render on my fork:
https://github.com/ens-lgil/ensembl-vep/blob/feature/readme_tags/README.md

PR #352, opened 2018-12-13
Grant all permissions on $OPT/src/ensembl-vep

alanhoyle:
Changing permissions on the whole $OPT/src/ensembl-vep directory tree allows the `perl INSTALL.pl` to run as users injected into the Docker container.

We run docker in an environment where we inject the user's ID into the container at runtime. This allows containers running on our cluster to access files that are otherwise inaccessible due to rootsquashing. When `perl INSTALL.pl` is run on the ensemblorg/ensembl-vep image, it fails out almost instantly with the following error:

```
Checking for installed versions of the Ensembl API...No such file or directory at INSTALL.pl line 692.
```

However, if the permissions are changed to a+rwx for the $OPT/src/ensembl-vep directory tree, INSTALL.pl works. This patch merely forces the below command to be run in docker build process.

(Note:  our user injection works automatically and is provided by AquaSec.  The transcript below is mildly edited)

See below for a demo:
```
$ docker run --rm -ti ensemblorg/ensembl-vep
alanhoyle@4023c784a4d2:/opt/vep/src/ensembl-vep$ perl INSTALL.pl
WARNING: Unable to carry out version check for ensembl-vep

Hello! This installer is configured to install v94 of the Ensembl API for use by the VEP.
It will not affect any existing installations of the Ensembl API that you may have.

It will also download and install cache files from Ensembl's FTP server.

Checking for installed versions of the Ensembl API...No such file or directory at INSTALL.pl line 692.
alanhoyle@4023c784a4d2:/opt/vep/src/ensembl-vep$ su vep -c "chmod -R a+rwx $PWD"
alanhoyle@4023c784a4d2:/opt/vep/src/ensembl-vep$ perl INSTALL.pl

Hello! This installer is configured to install v94 of the Ensembl API for use by the VEP.
It will not affect any existing installations of the Ensembl API that you may have.

It will also download and install cache files from Ensembl's FTP server.

Checking for installed versions of the Ensembl API...done
It looks like you already have v94 of the API installed.
You shouldn't need to install the API

Skip to the next step (n) to install cache files

Do you want to continue installing the API (y/n)? 
[...]
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -102,5 +102,5 @@ ENV PATH $OPT/src/ensembl-vep:$PATH
 
 # run INSTALL.pl
 WORKDIR $OPT/src/ensembl-vep
-RUN chmod u+x *.pl
+RUN chmod -R a+rwx .
 RUN ./INSTALL.pl -a a -l


alanhoyle:
Note:  a work-around to get the INSTALL.pl to work in the container as it exists today is to run the INSTALL.pl this way:  

```
 $ docker run --rm -ti -v $PATH_TO_CACHE ensemblorg/ensembl-vep bash -c "su vep -c 'chmod -R a+rwx /opt/vep/src/ensembl-vep' ; perl INSTALL.pl -c $PATH_TO_CACHE"

Hello! This installer is configured to install v94 of the Ensembl API for use by the VEP.
It will not affect any existing installations of the Ensembl API that you may have.

It will also download and install cache files from Ensembl's FTP server.

Checking for installed versions of the Ensembl API...done
It looks like you already have v94 of the API installed.
```

ens-lgil:
Dear @alanhoyle,

Thank you for the code suggestions.
I am glad you found a work around method to bypass the current Docker VEP image to suit your particular use case.
For security reasons, I am not entirely sure that you should grant write access in your change (i.e. `chmod -R a+rx`): 
`/opt/vep/src/ensembl-vep` already contains the VEP and others Ensembl APIs, therefore you shouldn't need write access here (unless you want to update the API version, which can be done by pulling a newer Docker image).
The only things that can be downloaded/installed are the cache files and the plugins. They should be installed (throught the Docker container) in an external directory.
The default use case is to mount a volume to the Docker container, pointing to a default directory (`/opt/vep/.vep`) and then install the cache in it, e.g. in our [documentation](https://www.ensembl.org/info/docs/tools/vep/script/vep_download.html#docker):
```
docker run -t -i -v $HOME/vep_data:/opt/vep/.vep ensemblorg/ensembl-vep
```

Best regards,
Laurent

alanhoyle:
Perhaps a better solution would be to check those permissions only after the "y" has been selected rather than crashing at that time?

aparton:
Hi,

Thank you for your interest in VEP, and for your contribution to our Docker image. After discussing it with the team, we have decided not to update the permissions within $OPT/src/ensembl-vep, as this folder only contains ensembl-vep and the required Ensembl API dependencies, so write access shouldn’t be required. For your particular use case, I recognise the benefit, but we don’t think it’s a change that we should include within our standard docker image.

If you have any other suggested changes to VEP, please feel free to create a pull request, or to contact us.

Kind Regards,
Andrew

PR #354, opened 2018-12-19
Fix/add missing funcgen patches 95

at7:


helensch:
Merged into master and release/95

PR #357, opened 2018-12-20
Update the VEP version for release 96

helensch:


helensch:
Merged into master and release/96

PR #359, opened 2019-01-03
SPDI reference sequence include MT

dglemos:


modules/Bio/EnsEMBL/VEP/Parser/SPDI.pm
```diff
@@ -185,7 +185,7 @@ sub create_VariationFeatures {
   }
 
   # throws error if reference sequence is not genomic
-  if($spdi !~ /^(NC\_|CHR|([0-9]{1,2}\:|X\:|Y\:))/i) {
+  if($spdi !~ /^(NC\_|CHR|MT|([0-9]{1,2}\:|X\:|Y\:))/i) {
```
helensch: Should MT be treated the same as X and Y i.e. with following colon
Not sure why ([0-9]{1,2}\:|X\:|Y\:) is bracketed 
Could maybe simplify regular expression e.g.
/^(NC\_|CHR|([0-9]{1,2}|X|Y|MT)\:)/i
Does allow for leading 0 and only up to 99, was this intended?

dglemos: > Does allow for leading 0 and only up to 99, was this intended?
yes, it's supposed to be the chromosome number so it only can have two digits.

PR #361, opened 2019-01-07
Update/improve Docker image

ens-lgil:
Minor improvements about the Docker image, still based on Ubuntu 16.04


ens-lgil:
This update is preparing the migration to Ubuntu 18.04 (which has a slightly smaller image and more recent libraries)

aparton:
When I build an image with this DockerFile, everything runs and completes happily, but I get a couple of warning messages:

- debconf: delaying package configuration, since apt-utils is not installed
- Warning: prerequisite Test::Deep 0 not found.
- Warning: prerequisite Test::LeakTrace 0 not found.

Are any of these packages particularly bloatworthy and worth not putting in, or should we add them?

ens-lgil:
The `apt-utils` warning is not really relevant according to some quick search on google.
For the others warnings I think these are dependencies from the cpanfile imports but I don't think they are critical (Test modules that we don't use if I am correct)

ens-lgil:
PR #376 is more up to date

PR #365, opened 2019-01-16
add cpanm requirement for G2P plugin

at7:
Add perl module requirement for G2P plugin to cpanfile


ens-lgil:
Merged into `release/96` and `master`

PR #372, opened 2019-01-30
Update bam file

at7:
switch to new bam file supporting MANE set


at7:
Assembly version mismatch. I will try again.

PR #373, opened 2019-01-30
update GRCH38 bam file

at7:
Use newly provided bam file supporting MANE set


ens-lgil:
Merged into `master` and `release/96`

PR #379, opened 2019-02-05
Remove README section comparing 'ensembl-vep' with 'ensembl-tools VEP'

ens-lgil:


at7:
merged to release/96 and master

PR #382, opened 2019-02-06
Remove some flags already deprecated and update the list of incompati…

ens-lgil:
…ble flags


at7:
merged to master and release/96

PR #385, opened 2019-02-12
Fix installation script when an older version of VEP is specified wit…

ens-lgil:
…h the flag '--VERSION'


ens-lgil:
I will create a new PR with a warning message

PR #386, opened 2019-02-12
Fix installation script when an older version of VEP is specified wit…

ens-lgil:
…h the flag '--VERSION'


at7:
Do we need to return an error message that the versions don't match?

ens-lgil:
This doesn't require an error message: the code change skips the unecessary check for the latest VEP version/subversion available when we use the flag `--VERSION`. This flag is used to install a specific version of VEP (the specified version ends up in `$API_VERSION`)

ens-lgil:
I will create a new PR deprecating this VERSION flag

PR #387, opened 2019-02-13
Remove the extra 'Extra' key in the default VEP output (found with th…

ens-lgil:
…e G2P plugin)

e.g. `Extra=HASH(0x2819cdf0)`:
```
rs80338647	1:235806723	-	ENSG00000143669	ENST00000389793	Transcript	frameshift_variant	2588	2413	805	E/X	Gaa/aa	rs80338647,CD075459,COSM210418	IND=patient;ZYG=HOM;IMPACT=HIGH;STRAND=-1;SYMBOL=LYST;SYMBOL_SOURCE=HGNC;HGNC_ID=HGNC:1968;CLIN_SIG=pathogenic;SOMATIC=0,0,1;PHENO=1,1,1;Extra=HASH(0x2819cdf0);G2P_complete=1;G2P_flag=HOM;G2P_gene_req=BI
```

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory/VEP_output.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory/VEP_output.pm
@@ -106,7 +106,7 @@ sub output_hash_to_line {
   # add additional fields to "Extra" col at the end
   my %extra =
     map {$_ => $hash->{$_}}
-    grep {!$OUTPUT_COLS_HASH{$_}}
+    grep {!$OUTPUT_COLS_HASH{$_} && $_ ne 'Extra'}
     keys %$hash;
 
   my $field_order = $self->field_order;


at7:
Do we need to return an error message that the versions don't match?

at7:
Could there not be cases when Extra should be added to the output? It looks like Extra hash is not dereferenced correctly. But I still don't know why it is added in the first place. Thanks for looking into this.

ens-lgil:
Yes, this is the risk ... however I couldn't find where/how it is added in the code.

at7:
I have same time after handover to take a look!

aparton:
I believe it's added in the G2P plugin at:   `my $zyg = $line->{Extra}->{ZYG} || $line->{ZYG};` (line 395 on feature/g2p_af_no_dbc). Should be as simple as changing it to 
`my $zyg = defined($line->{Extra}) ? $line->{Extra}->{ZYG} : $line->{ZYG};`
but I haven't tested it thouroughly

at7:
Thank you @aparton It never occurred to me that Extra is not always defined! What an improvement. I'll fix it for my PR.

at7:
@ens-lgil Thank you for the suggested fix. But I think it is better not to add an empty extra hash in the G2P plugin in the first place. I suggest that this PR is closed without merge.

ens-lgil:
Ok no problem, as I said in Andrew's PR, this wasn't the ideal fix.

PR #392, opened 2019-02-15
Bugfix/validate vf by allele length

at7:
Web team reported a VEP issue caused by variants whose reference allele were thousands of bp long. I set a threshold of 50 bp for input allele length before the variant is failing the validation step.


at7:
@ens-lgil Do we need to increment VEP_SUB_VERSION?

ens-lgil:
Yes we will need to update the subversion.
However there is an other PR pending at the moment with the subversion `2`

at7:
I will do more tests to determine the right allele length for setting the threshold.

at7:
Some  more [investigation](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1642) required to define the right threshold

PR #400, opened 2019-02-19
3' shifting

aparton:
Linked to ensembl-variation PR: https://github.com/Ensembl/ensembl-variation/pull/314

PR #404, opened 2019-02-25
Automatically shift insertions/deletions within repeated regions to their most 3' position

aparton:
Rebased version of #400


aparton:
I've updated the travis.yml file - don't merge this until it's been restored.

.travis.yml
```diff
@@ -35,7 +35,7 @@ before_install:
   - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl.git
   - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-io.git
   - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-funcgen.git
-  - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-variation.git
+  - git clone --branch feature/3prime-shifting-master --depth 1 https://github.com/aparton/ensembl-variation.git
```
ens-lgil: Dev configuration to change before merging

modules/Bio/EnsEMBL/VEP/AnnotationType/Variation.pm
```diff
@@ -84,9 +84,9 @@ sub annotate_InputBuffer {
     foreach my $vf(
       grep {ref($_) ne 'Bio::EnsEMBL::Variation::StructuralVariationFeature'}
       @{$buffer->get_overlapping_vfs($existing_vf->{start}, $existing_vf->{end})}
-    ) {
+    ) {      
       my $matched = $self->compare_existing($vf, $existing_vf);
-      push @{$vf->{existing}}, $matched if $matched;
+      push @{$vf->{existing}}, $matched if ($matched && !(grep($_->{variation_name} eq $matched->{variation_name},@{$vf->{existing}})));
```
sarahhunt: There is no point checking if a variant matches if the result will be thrown away if a match is already known.

Where do the duplicates come from - may they have different matched_alleles which may impact the check in add_colocated_variant_info?

t/Runner.t
```diff
@@ -217,8 +217,11 @@ is_deeply($runner->get_OutputFactory, bless( {
   'cell_type' => undef,
   'mirna' => undef,
   'ambiguity' => undef,
+  'shift_genomic' => undef,
+  'no_shift' => 0,
   'vcf_string' => undef,
   'clin_sig_allele' => undef,
+  'shift_genomic' => undef,
 }, 'Bio::EnsEMBL::VEP::OutputFactory::VEP_output' ), 'get_OutputFactory');
```
sarahhunt: Why is shift_genomic repeated?

modules/Bio/EnsEMBL/VEP/AnnotationType/Variation.pm
aparton: Checking is done on shifted and unshifted values, so if you have a colocated variant inside a deletion that's shifted, then it may appear more than once

We shouldn't get different matched alleles as they're coming from the $vf object rather than the $vfoa object

aparton:
Squashing and merging. Failing test is due to regulation update

PR #405, opened 2019-03-07
Fix Ensembl APIs installation issue when target directory already exi…

ens-lgil:
…st in ensembl-vep/Bio/EnsEMBL

For example `ensembl-io` has a new directory `Utils` ([ensembl-io/modules/Bio/EnsEMBL/Utils](https://github.com/Ensembl/ensembl-io/tree/release/96/modules/Bio/EnsEMBL)) and `ensembl` has also a directory `Utils` ([ensembl/modules/Bio/EnsEMBL/Utils](https://github.com/Ensembl/ensembl/tree/release/96/modules/Bio/EnsEMBL)).

Both directory are moved into `ensembl-vep/Bio/EnsEMBL/Utils` by the INSTALL script.
However it failed because the `move()` method can't overwrite an existing directory: ensembl/modules/Bio/EnsEMBL/Utils is moved first and it failed when the INSTALL script tries to move ensembl-io/modules/Bio/EnsEMBL/Utils in `ensembl-vep/Bio/EnsEMBL/`.

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -132,10 +132,10 @@ $| = 1;
 
 # other global data
 my @API_MODULES = (
-  { name => 'ensembl',           path => '',          test_pm => 'Bio::EnsEMBL::Registry' },
+  { name => 'ensembl',           path => ' ',         test_pm => 'Bio::EnsEMBL::Registry' },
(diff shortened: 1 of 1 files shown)


aparton:
I'm happy with this. Passes installation tests on Ubuntu 16.04 and Mojave. Travis tests fail due to so_term issue, but that's not related to this PR.

PR #409, opened 2019-03-08
ignore empty motifFeatures

at7:
This fix will ignore empty MotifFeatures. For release/96 we are more carful with dumping regulatory data into cache files and won't have to deal with reading empty MotifFeatures from cache files.


ens-lgil:
We need to upgrade the VEP subversion as well.

PR #418, opened 2019-03-21
ancestral_allele patch

at7:
Patch test db with ancestral allele change.


ens-lgil:
This is pretty much what I did in the PR https://github.com/Ensembl/ensembl-vep/pull/416
Maybe I missed something or transpose the wrong ancestral alleles to `variation_feature`.

at7:
Sorry, we have to move the ancestral alleles from variation to variation_feature. I didn't include the join into the PR because this would have killed staging for human.

PR #434, opened 2019-04-12
Haplo REST docs

sarahhunt:
Brief description of JSON output format


ens-lgil:
I just updated my sandbox with the updated doc:
http://ves-hx2-76.ebi.ac.uk:5060/info/docs/tools/vep/haplo/index.html#hp_rest

Issue #459, opened 2019-05-22
Haplosaurus: contributing variants from different chromosomes


Shakamaka:
I have noticed strange behaviour with Haplosaurus and how it handles variants across chromosomes. I have created a minimal example reproducing the strangeness.
Running:
`haplo --assembly GRCh38 --dir /home/databases/variant_effect_predictor/.vep --offline -i test2.vcf -o STDOUT`
Gives:
![image](https://user-images.githubusercontent.com/50915036/58176019-b46afc80-7ca1-11e9-99b6-cd594f2c6217.png)

With the following, simplified test2.vcf file:

> #CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	T1
6	142134009	testChr6	C	A	.	PASS	DP=289	GT	0|1
7	142133999	testChr7	G	A	.	PASS	DP=760	GT	0|1

As you see, Haplosaurus adds variants from two different chromosomes (chr6 and chr7) onto the same protein haplotype. I have noticed many such examples in my data, and, worryingly, it seems to happen for variants with a very similar genomic position (yet different chromosomes).
What is the reason for this behaviour?

![image](https://user-images.githubusercontent.com/50915036/58177113-0745b380-7ca4-11e9-812e-8ad049ba4c3c.png)

ens-lgil:
Dear @Shakamaka,

You are right: this behaviour is not expected by Haplosaurus.
I believe this is due to some optimisations in the code which are not expecting 2 following entries with similar coordinates but on different chromosomes.

We are investigating the issue and we will get back to you when the issue will be fixed.

Best regards,
Laurent

ens-lgil:
Dear @Shakamaka,

We identified the issue and added a fix on the branch `release/96`.
Please, could you update your VEP installation to get the fixed code ?

Best regards,
Laurent

Shakamaka:
Hi. I have tested the new release, and it works on the minimal example that I provided. However, this only seems to work when the chromosomes are given as numbers. If I use the following VCF:
> #CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	T1
chr6	142134009	testChr6	C	A	.	PASS	DP=289	GT	0|1
chr7	142133999	testChr7	G	A	.	PASS	DP=760	GT	0|1

I get this:
![image](https://user-images.githubusercontent.com/50915036/59018402-1b61e700-8846-11e9-9b8f-8fe8183afeb7.png)

[...]

PR #462, opened 2019-05-23
Fix interpretation issue when following entries have similar coordina…

ens-lgil:
…tes but on different chromosomes in the input file


ens-lgil:
Related to the issue https://github.com/Ensembl/ensembl-vep/issues/459

ens-lgil:
Cleaner PR here: https://github.com/Ensembl/ensembl-vep/pull/471

PR #463, opened 2019-05-23
skip long and complex SV and enable overlaps in VCF

sarahhunt:
Complex SV cannot currently be annotated, so skip them. 
Long SV require large amounts of memory, so a default max size is set which can be overridden on the command line.
Docs to follow


ens-lgil:

PR #465, opened 2019-05-24
Variant Recoder - return vcf format 

dglemos:
Return vcf format '1-55516888-G-GA' (optional field)


ens-lgil:
Seems to be duplicated with https://github.com/Ensembl/ensembl-vep/pull/470.
Should we close it ?

dglemos:
Closing it.

Issue #438, opened 2019-04-17
Confirming VCF REF allele matches reference sequence with local RefSeq cache


danhovelson:
I'm running VEP using a local Refseq cache (offline) [VEP Docker image v95.3], and would like to systematically confirm that the REF allele provided in my submitted VCF matches that present in my provided FASTA file for all variants submitted for annotation. 

I'm wondering why (when using `--check_ref` with `--dont_skip`) there's no indication that the REF allele in `my.vcf` **does not match** the REF allele in the provided FASTA file.

Example full VEP command:

```bash
cat my.vcf | vep \
        -o STDOUT \
        --offline \
        --refseq \
        --format vcf \
        --fasta $FASTA_FILE \  # note: here we're using hg19.fasta
        --use_given_ref \
        --check_ref \
        --dont_skip \
        --hgvs \
        --force_overwrite \
        --exclude_predicted \
        --symbol \
        --vcf \
        --flag_pick_allele \
        --no_escape \
        --canonical \
        --allele_number \
        --fork 4
```

Contents of `my.vcf` file:
```bash
##fileformat=VCFv4.1
##reference=grch37
##sampleGender=Unknown
##contig=<ID=chr1,length=249250621,assembly=hg19>
##contig=<ID=chr2,length=243199373,assembly=hg19>
##contig=<ID=chr3,length=198022430,assembly=hg19>
##contig=<ID=chr4,length=191154276,assembly=hg19>
##contig=<ID=chr5,length=180915260,assembly=hg19>
##contig=<ID=chr6,length=171115067,assembly=hg19>
##contig=<ID=chr7,length=159138663,assembly=hg19>
##contig=<ID=chr8,length=146364022,assembly=hg19>
##contig=<ID=chr9,length=141213431,assembly=hg19>
##contig=<ID=chr10,length=135534747,assembly=hg19>
##contig=<ID=chr11,length=135006516,assembly=hg19>
##contig=<ID=chr12,length=133851895,assembly=hg19>
##contig=<ID=chr13,length=115169878,assembly=hg19>
##contig=<ID=chr14,length=107349540,assembly=hg19>
##contig=<ID=chr15,length=102531392,assembly=hg19>
##contig=<ID=chr16,length=90354753,assembly=hg19>
##contig=<ID=chr17,length=81195210,assembly=hg19>
##contig=<ID=chr18,length=78077248,assembly=hg19>
##contig=<ID=chr19,length=59128983,assembly=hg19>
[...]

PR #469, opened 2019-05-28
Adds check_ref warning to output file

aparton:
A solution to: https://github.com/Ensembl/ensembl-vep/issues/438


modules/Bio/EnsEMBL/VEP/Constants.pm
```diff
@@ -119,6 +119,7 @@ our @FLAG_FIELDS = (
   { flag => 'check_existing',  fields => ['CLIN_SIG','SOMATIC','PHENO'] },
   { flag => 'pubmed',          fields => ['PUBMED'] },
   { flag => 'check_svs',       fields => ['SV'] },
+  { flag => 'dont_skip',       fields => ['CHECK_REF'] },
```
ens-lgil: Not sure about the flag name: it looks too generic for me.

modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -646,6 +646,7 @@ sub validate_vf {
     }
 
     if(!$ok) {
+      $vf->{check_ref_failed} = 1;
```
ens-lgil: Could we indicate the input line (e.g. ID, chr, position) in the warning message ?

aparton: All this is doing is adding the failed status to the VEP output, which already contains the input position.

modules/Bio/EnsEMBL/VEP/Constants.pm
aparton: Fair enough, what do you suggest? FAILED_CHECK_REF?

I didn't want to add any data for when it's fine, mainly because it would normally just bloat the output file with extra text. The default behaviour is to skip variants that don't match, this flag will only appear by forcing the inclusion of non-matching variants with --dont_skip.

And I don't think we need to add the flag into the `vep` flags list, `--dont_skip` already exists, I'm just appending something new to the output when it's used

ens-lgil: Fair enough, I guess this is ok then

ens-lgil:

PR #470, opened 2019-05-28
Variant recoder - return vcf format e97 

dglemos:
Duplicated #465 (e98)  
variant recoder returns VCF format (optional)


modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -839,6 +840,25 @@ sub VariationFeature_to_output_hash {
     Location            => ($vf->{chr} || $vf->seq_region_name).':'.format_coords($vf->{start}, $vf->{end}),
   };
 
+  my $converted_to_vcf = $vf->to_VCF_record;
+
+  my $alt_allele_vcf = ${$converted_to_vcf}[4];
+
+  if($self->{vcf_string} || grep(/vcf_string/, @{$self->{_config}->{_params}->{fields}})){
```
sarahhunt: Only $self->{vcf_string} is needed/ useful here

sarahhunt:
This is working nicely, but the results are not returned left shifted (eg. ENST00000361337.3:c.-234del) can you see if this can be done efficiently? We'd need to check prev_base here:

https://github.com/Ensembl/ensembl-variation/blob/release/96/modules/Bio/EnsEMBL/Variation/VariationFeature.pm#L2304

PR #478, opened 2019-06-04
Allow the use of ESP as frequency group

ens-lgil:


modules/Bio/EnsEMBL/VEP/AnnotationSource/Cache/BaseCacheVariation.pm
```diff
@@ -271,7 +271,8 @@ sub check_frequency_filter {
   throw("ERROR: Invalid population ".$self->{freq_pop}." specified")
     unless $freq_pop_full eq '1KG_ALL' ||
     $valid{$freq_pop_full} ||
-    ($freq_group eq '1KG' && $valid{$freq_pop || ''});
+    ($freq_group eq '1KG' && $valid{$freq_pop || ''}) ||
+    ($freq_group eq 'ESP' && $valid{$freq_pop || ''});
```
helensch: Could a comment be added about why would have $valid{''}  - i.e. empty key
Is there a test to check ESP filtering before and after this change

at7: @ens-lgil could you please add the test for ESP filtering?

ens-lgil:
This might be something to rewrite regarding some discussions last week with Andrew

PR #482, opened 2019-06-05
Docs - update Haplosaurus to specify input data must be sorted

dglemos:


dglemos:
It would be better for 97. I'm closing this one then.

PR #487, opened 2019-06-06
Haplo - fix chromosome match 

dglemos:


helensch:
Suggest adding a check for using chr1 instead of 1

dglemos:
Thanks Helen!

PR #488, opened 2019-06-06
Haplo - fix chromosome match (97) 

dglemos:


ens-lgil:
Merged into `release/97` and `master`

PR #489, opened 2019-06-10
Update the error message when an input file is not sorted by location

ens-lgil:


ens-lgil:

PR #492, opened 2019-06-12
Haplo - check if chromosome is valid 

dglemos:


modules/Bio/EnsEMBL/VEP/Haplo/Parser/VCF.pm
```diff
@@ -191,4 +193,35 @@ sub create_VariationFeatures {
   }];
 }
 
+=head2 validade_chr
```
helensch: Type in header

modules/Bio/EnsEMBL/VEP/Haplo/Parser/VCF.pm
```diff
@@ -191,4 +193,35 @@ sub create_VariationFeatures {
...
+
+  Arg 1      : hash
+  Example    : my $vf_hashes = $parser->validade_chr($vf);
```
helensch: Typo in example

modules/Bio/EnsEMBL/VEP/Haplo/Parser/VCF.pm
dglemos: That's right. Thanks.

ens-lgil:
A general question: would it be useful for VEP as well ?
I mean, should we move the new method `validate_chr` in `Bio::EnsEMBL::VEP::Parser` or `Bio::EnsEMBL::VEP::Parser::VCF` ?

at7:
merged with master and release/97

PR #496, opened 2019-06-14
Bugfix/resume after homref

at7:
A G2P collaborator noticed that VEP prematurely stops if the last variant on a chromosome is homozygous for the reference or not present in all individuals and the next variant is on a different chromosome. This only occurs for tab and not vcf output. If you run the updated test without my added line in Runner.pm it should fail, adding the line is fixing the test.


at7:
I have removed it with `remove debug module` commit. Are you looking at the latest version?

helensch:
Sorry, must have been looking at older version.

t/Runner.t
```diff
@@ -1023,6 +1022,23 @@ STACK S3
...
+$runner = Bio::EnsEMBL::VEP::Runner->new({
+  %$cfg_hash,
+  input_data => undef,
+  input_file => $test_cfg->{test_vcf5},
+  dir => $test_cfg->{cache_root_dir},
+  output_file => $test_cfg->{user_file}.'.out',
+  individual => 'all',
+});
+ok($runner->run, 'run - ok');
+open IN, $test_cfg->{user_file}.'.out';
```
helensch: $test_cfg->{user_file}. '.out' should be unlinked when finished testing.

at7:
@helensch Can you please try again with the new test5.vcf file? We can go through the code changes in our meeting.

helensch:
Thanks for the adding the test for a variant with REF==ALT. This also fixes the problem that VEP was stopping when the last variant on chromosome had REF==ALT, no genotypes and output was tab.

t/testdata/input/test6.vcf
```diff
@@ -0,0 +1,4 @@
+##fileformat=VCFv4.1
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
+21	25585733	rs142513484	C	c	.	.	.
```
helensch: Suggest making ALT a C instead of C

Issue #453, opened 2019-05-17
vep93 docker randomly stops, hangs, freezes


thedam:
Hey, I always run the same command on my VCF files that have always 100 variants.

2 out of 10 times the VEP just hangs. I just see endlessly: 
`2019-05-17 10:52:50 - INFO: BAM-edited cache detected, enabling --use_transcript_ref; use --use_given_ref to override this` and it never finishes. I have to CTRL + C the process and re-run exactly the same command, so then it usually works.

Any idea why it happens, how I can debug it or see some logs? 
Thanks

```
docker run --rm -t -i \
-v /mnt/ssd_01/VEP/vepCache:/opt/vep/.vep \
-v /mnt/ssd_01/DATA:/home/vep/DATA  \
-v /mnt/sata_03/DATABASES:/home/vep/DATABASES \
-v $PWD:$PWD \
-v /mnt/ssd_01/refs/:/home/vep/refs ensemblorg/ensembl-vep vep \
--format "vcf" \
-i $inFileFull \
-o $outVepFile \
--no_stats \
--cache \
--fasta /home/vep/refs/hs37d5_noHap.fa --port 3337 --ASSEMBLY GRCh37 \
--merged --tab \
--use_transcript_ref \
--variant_class --sift b --polyphen b --humdiv --gene_phenotype --regulatory \
--hgvs --hgvsg --shift_hgvs 1 --symbol --protein --ccds --uniprot  --numbers --domains --canonical --biotype \
--af --af_1kg --af_esp --af_gnomad --max_af --pubmed --force_overwrite \
--exclude_predicted \
--buffer_size 200 \
--plugin dbscSNV,/home/vep/DATA/dbscSNP/dbscSNV1.1_GRCh37.txt.gz  \
--plugin SpliceRegion \
--plugin MaxEntScan,/home/vep/DATA/MaxEntScan \
--plugin GeneSplicer,/home/vep/DATA/GeneSplicer/sources/genesplicer,/home/vep/DATA/GeneSplicer/human \
--plugin ExACpLI,/home/vep/DATA/ExACpLI/ExACpLI_values.txt \
--plugin Phenotypes \
[...]

PR #504, opened 2019-06-17
mv waring message

at7:
Moving warning message into the right place https://github.com/Ensembl/ensembl-vep/issues/453

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser.pm
@@ -137,7 +137,7 @@ sub new {
     # detect format
     if(lc($format) eq 'guess' || lc($format) eq 'detect' || lc($format) eq 'auto') {
       $format = $self->detect_format();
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -137,15 +137,14 @@ sub new {
     # detect format
     if(lc($format) eq 'guess' || lc($format) eq 'detect' || lc($format) eq 'auto') {
       $format = $self->detect_format();
+      $self->warning_msg("No input file format specified - detected $format format") if $self->param('verbose');
```
ens-lgil: Might also check if a `format` is returned` ?
e.g. ` ... if ($self->param('verbose') &&  $format);

at7: https://github.com/Ensembl/ensembl-vep/blob/d68cab0666bfe36b7d56dc50fa95806f1eda3fb9/modules/Bio/EnsEMBL/VEP/Parser.pm#L143

PR #507, opened 2019-06-18
patch test dbs for release/98

at7:


at7:
merged to master and release/98

PR #508, opened 2019-06-18

aparton:
Still sorting out getting the CHECKSUMS file updated on the FTP site, please don't merge until that's sorted

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -56,6 +56,7 @@ our (
   $ENS_GIT_ROOT,
   $BIOPERL_URL,
   $CACHE_URL,
+  $CACHE_URL_INDEXED,
   $CACHE_DIR,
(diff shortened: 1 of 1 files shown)


INSTALL.pl
```diff
@@ -1086,7 +1093,7 @@ ()
     push @files, grep {$_ =~ /tar.gz/} $ftp->ls;
   }
   else {
-    opendir DIR, $CACHE_URL;
+    opendir DIR, $URL_TO_USE;
```
at7: can we call it $CACHE_URL_TO_USE. It will be more readable for the remaining of the script

ens-lgil:
We will need to create a similar PR against `postreleasefix/98`

ens-lgil:
Merged to `release/97`

PR #512, opened 2019-06-20
Add buffer_size check to sort command in dumping pipeline

aparton:
Prevents the issues we had in 97 with the GRCh37 all_vars.gz files being empty after dumping


at7:

PR #516, opened 2019-06-26
Fixes rest test failures

aparton:
Moves check for clinvar source_id to before the variation_feature query


ens-lgil:

PR #517, opened 2019-06-27
Minor updates, including comments

ens-lgil:


ens-lgil:
Needs to do some changes regarding the new htslib and Bio::DB::HTS supported versions and waiting for the relevant PRs to be merged to VEP before committing the Dockerfile change.

PR #523, opened 2019-07-03
Variant recoder - update README with spdi and vcf_string info

dglemos:


ens-lgil:
We will need to create a PR against `postreleasefix/98` as well

dglemos:
Wil do

PR #524, opened 2019-07-03
Variant recoder - update docs with vcf_string field

dglemos:


ens-lgil:
Merged into `release/98` and `master`

Issue #530, opened 2019-07-12
VEP97 Unsorted VCF error on sorted vcf.  


thomasperson:
Howdy,

When running VEP97 with the command:
vep -i gnomad.exomes.r2.1.1.sites.21.sorted.vcf -o gnomad.exomes.r2.1.1.sites.21.VEP97.vcf --everything --cache --offline --vcf --dir_cache CACHEDIR/ --fasta human_g1k_v37_decoy.fasta.gz --merged --no_stats --force_overwrite --assembly GRCh37 -v --use_given_ref --format vcf

It is exiting with the error: 
Exiting the program. The input file appears to be unsorted. Please sort by chromosome and by location and re-submit.

I downloaded the sites only exome vcf from https://gnomad.broadinstitute.org/downloads.  I also uncompressed it and sorted with vcf-sort from vcftools to ensure it is sorted and used the b37 fasta is from the GATK 2.8 bundle.  gnomad.exomes.r2.1.1.sites.Y.vcf.bgz  works with no issue.  

Don't know if its relevent, but the downloaded cache files for homo_sapiens_merged_vep_97_GRCh37.tar.gz are 7GB smaller than they were for v96. I deleted them and redownloaded and still smaller and still same issues.

Thanks for your help,

Nate

ens-lgil:
Dear @thomasperson,

I am investigating the gnomAD: file https://storage.googleapis.com/gnomad-public/release/2.1.1/vcf/exomes/gnomad.exomes.r2.1.1.sites.21.vcf.bgz

We implemented the sorting check as unsorted VCF files slow down VEP and use more memory.
In the meantime, you can add the option [--no_check_variants_order](https://www.ensembl.org/info/docs/tools/vep/script/vep_options.html#opt_no_check_variants_order) to skip the sorting check.

About the VEP cache size differences, I can only see a big difference for the default VEP indexed cache file:
- 20Gb: ftp://ftp.ensembl.org/pub/release-96/variation/indexed_vep_cache/homo_sapiens_vep_96_GRCh37.tar.gz
- 13Gb: ftp://ftp.ensembl.org/pub/release-97/variation/indexed_vep_cache/homo_sapiens_vep_97_GRCh37.tar.gz

Is this actually that file you a refering ?
I downloaded the `96` and `97` `merged` files and they are all between 13Gb and 14Gb.
What it the size of your downloaded 97 cache file ?

Best regards,
Laurent

seru71:
[...]

PR #532, opened 2019-07-15
Use a default distance to check if the variant is still in the same r…

ens-lgil:
…egion even if it's not ordered with the previous variant location: we use the VF start and not the VCF start (they can be different when the variant is a deletion for instance and/or when the alleles can be minimised)


ens-lgil:
Proposed fix for the reported issue https://github.com/Ensembl/ensembl-vep/issues/530

at7:
Can you please merge the changes into master and release/98

PR #539, opened 2019-07-23
Prevent sorting when using list of rsIDs

aparton:


at7:
merged to master and release/98

PR #541, opened 2019-07-24
HTSLib & Bio::DB::HTS updates within installer

aparton:
Linked to https://github.com/Ensembl/ensembl-vep/pull/536

Updates the versions of HTSLib and Bio::DB::HTS within the VEP installer. New dependancies are introduced that the installer asks the user to add.

A discussion needs to be had within the team of the best way to resolve these dependancies across multiple operating systems.

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -802,36 +802,46 @@ END
 
   apt-get install build-essential
 END
-
+  my $msg = '';
   my $this_os =  $^O;
   if( $this_os ne 'darwin' ) {
-    -e '/usr/include/zlib.h' or die <<END;
-      zlib.h library header not found in /usr/include. Please install it and try again.
+    unless(-e '/usr/include2/zlib.h'){
+    
+      $msg += 'zlib.h library header not found in /usr/include. Please install it and try again.
       (or to skip Bio::DB::HTS/htslib install re-run with --NO_HTSLIB)
 
       On Debian/Ubuntu systems you can do this with the command:
 
       apt-get install zlib1g-dev
-END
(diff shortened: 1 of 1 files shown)


INSTALL.pl
```diff
@@ -802,36 +802,46 @@ END
 
   apt-get install build-essential
 END
-
+  my $msg = '';
   my $this_os =  $^O;
   if( $this_os ne 'darwin' ) {
-    -e '/usr/include/zlib.h' or die <<END;
-      zlib.h library header not found in /usr/include. Please install it and try again.
+    unless(-e '/usr/include2/zlib.h'){
```
ens-lgil: Could I suggest something like this to make the code more compact ?

```
my $this_os =  $^O;
if( $this_os ne 'darwin' ) {

  my $msg = '';
 
  # List the required libraries with their packages
  my %libs = { 
    'zlib.h' => 'zlib1g-dev', 
    'lzma.h' => 'liblzma-dev', 
    'bzlib.h' => ''libbz2-dev'
  };

  my $default_msg = qq{%s library header not found in /usr/include. Please install it and try again.
(or to skip Bio::DB::HTS/htslib install re-run with --NO_HTSLIB)

On Debian/Ubuntu systems you can do this with the command:

apt-get install %s};

  # Loop over the required libraries
  foreach my $lib (sort(keys(%libs))) {
    unless(-e '/usr/include2/'.$lib){
      $msg += sprintf($default_msg, $lib, $libs{$lib));
    }
  }
  die($msg) if($msg ne '');
}
```

PR #551, opened 2019-07-29
Bugfix/sorting97

aparton:


modules/Bio/EnsEMBL/VEP/InputBuffer.pm
```diff
@@ -203,6 +203,13 @@ sub next {
       $prev_chr = $vf->{chr};
     }
   }
+  
+  my %unsorted_formats = (
```
helensch: Could this go in Config.pm?
Could it be quoted in same way as hashes in Config.pm?

t/InputBuffer.t
```diff
@@ -502,6 +502,19 @@ eval {
 };
 ok(!$@, "Skip the count for non ordered variants with the flag 'no_check_variants_order'");
 
+# Skip check for hgvs variants
+# To workaround the offline check for HGVS input, the parser being given to the InputBuffer is in VCF format. The sorting is correctly prevented by the InputBuffer, compared to the previous 2 tests.
```
helensch: Suggest splitting these comments onto shorter lines

PR #559, opened 2019-08-07
Update dumping pipeline to include MANE_select values

aparton:


ens-lgil:

PR #562, opened 2019-08-12
fix 3R chrom comp - ENSVAR-2196

ima23:


modules/Bio/EnsEMBL/VEP/Stats.pm
```diff
@@ -576,7 +576,7 @@ sub generate_chart_data {
     options => '{legend: {position: "none"}}',
   } if $stats->{chr_totals};
   
-  foreach my $chr(sort {($a !~ /^\d+$/ || $b !~ /^\d+/) ? $a cmp $b : $a <=> $b} keys %{$stats->{chr}}) {
+  foreach my $chr(sort {($a !~ /^\d+$/ || $b !~ /^\d+/ || $a =~ /^\d\w/ || $b =~ /^\d\w/ ) ? $a cmp $b : $a <=> $b} keys %{$stats->{chr}}) {
```
ens-lgil: Is this possible to have a chromosome like "11A"  (basically more than 1 number + 1 letter) ?
If so, we might need to use the regex `/^\d+\w/` instead

ima23: @ens-lgil: I'm not aware of a species with double digits followed by a letter. `drosophila melanogaster`, the source for this fix, has only one digit: 2L, 2R, 3L, 3R.

ens-lgil:
Merged into `release/98` and `master`

PR #567, opened 2019-08-19
Generate a file containing md5sums for all files within the cache

aparton:


at7:

PR #570, opened 2019-08-20
Update Stats.pm

epaule:
breaks on $b for names like chromosome "3R"

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Stats.pm
+++ b/modules/Bio/EnsEMBL/VEP/Stats.pm
@@ -576,7 +576,7 @@ sub generate_chart_data {
     options => '{legend: {position: "none"}}',
   } if $stats->{chr_totals};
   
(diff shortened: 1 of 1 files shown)


at7:
We recently fixed this for 98 https://github.com/Ensembl/ensembl-vep/pull/562/files I will close your PR and submit the same we also use for 98. I will also need to update the sub version. I let you know when the fix is in.

epaule:
thanks, my VEP checkout was from 7 days ago, so it didn't include https://github.com/Ensembl/ensembl-vep/pull/562/commits/f29209def4f1c7f4463ea4709abb31d8793c49a7

at7:
No problem. Are you working with release/97 or 98? Do you need the fix on release/97

epaule:
I used master of 7 days ago, but github defaulted to 97 when I did the pull request. I guess that is the default branch at the moment. So all good.

PR #584, opened 2019-08-30
Fix for regex matches against strings with odd characters

willmclaren:
"path" as a value would match "pathogenic&benign" but not
"benign&pathogenic" as the string was reduced to "benign" in an
attempt to separate text from numerical components.


aparton:
Hi Will, 

Thanks for this, any chance you could point this PR towards postreleasefix/98 instead?

Kind Regards,
Andrew

willmclaren:
Sure. It's been a while, I forgot the procedure :-)

PR #590, opened 2019-09-03
Revert md5 changes

aparton:
Causing a slowdown for production dumps - we'll need to change this for next time - potentially by only having this work for human.


at7:
merged into release/98 and master

PR #601, opened 2019-09-12
Fix HGVSp - can't uniquely determine nucleotide change

dglemos:


aparton:
With this fix then we still see the reverse issue with NMs being looked up in the core DB (i.e. http://rest.ensembl.org/vep/human/hgvs/NM_000527.4:p.M298L?content-type=application/json)

How do we want VEP to react in cases like this, do we want to just skip the variant and warn the user, or do we want to crash out entirely?

dglemos:
NM_000527.4:p.M298L doesn't work because it tries to fetch NM_000527.4 with fetch_by_stable_id(), here https://github.com/Ensembl/ensembl-variation/blob/7b96c960d2e462c8cf80ccb56914c199bf790fdf/modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm#L1876. I'll fix that.

aparton:
The fix is a definite improvement on what we have currently. I have a couple of questions though:

1) Are we happy with the error message that this gives when we can't uniquely get the nucleotide change from an HGVSp input?

_Unable to parse HGVS notation 'NM_000527.4:p.M298L': Can't call method "get_TranscriptMapper" on an undefined value at /nfs/public/release/ensweb/live/rest/www_98/ensembl-variation/modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm line 2323._

2) The change only looks at ENST transcripts and NMs, any other transcript type (like an XM) will still cause the initial issue. I'm not sure that pattern matching the stable_id is the best way of going about this. Can we check to see if the transcript exists in the core/otherfeatures database before running fetch_by_hgvs_notation? Would that slow things down too much?

3) Your coverage has decreased, can we add a test?

Issue #602, opened 2019-09-16
different  SIFT and CADD_PHRED values for the same variant depending on several factors


QANGFQ:
Hi, 

I found that SIFT values ​​are different for the same variant depending on output format(tab or vcf), --fields option and whether web version or command line version. 
The same is true for CADD_PHRED.

The command part of fields option was _--fields "SIFT, CADD_PHRED"_ and genome assembly version is GRCh37 and VEP command line version is 97.3.

Using rs199789235 as an example, the results are as follows.

 **tab & web version :** SIFT=-, CADD_PHRED=24.1
 **tab & command line version :** SIFT=0, CADD_PHRED=-
 **tab & command line version & fields option :** SIFT=0, CADD_PHRED=-
 **vcf & command line version & fields option :** SIFT=NULL, CADD_PHRED=NULL

I think that 'SIFT=0, CADD_PHRED=24.1' notation is correct. 
Is this right?
Why are the values different, despite the same variant?

Best regards,
QANGFQ

helensch:
Hi

SIFT predicts whether an amino acid substitution is likely to affect protein function.
https://www.ensembl.org/info/genome/variation/prediction/protein_function.html#sift

VEP reports SIFT by transcript and for missense variants.

The variant rs199789235 falls in more than one transcript.                      

For example 
- for transcript ENST00000327044, the consequence is downstream_gene_variant, there is no SIFT prediction and the value reported is "-'
- for transcript ENST00000341065, the consequence is missense_variant and the SIFT prediction is deleterious(0)

For command line VEP, to get CADD_PHRED, the CADD VEP plugin is needed:
https://www.ensembl.org/info/docs/tools/vep/script/vep_plugins.html
https://github.com/Ensembl/VEP_plugins/blob/release/97/CADD.pm

Helen

QANGFQ:
Hi, 
Thanks to your explanation, the issue related with CADD has been solved!
However, SIFT issue is not yet solved.. 

Transcript for the variant which is rs199789235 : chr1_874697_C>G I mentioned earlier is ENST00000342066.
That is, SIFT scores of ENST00000342066 are "-"(tab & web version), NULL(vcf & command line) and 0(tab & command line).

Why are the values differently marked, despite the same transcript?

Best regards,
QANGFQ

helensch:
Hi

Thank you for reporting this issue.

The SIFT score for rs199789235 on ENST00000342066 is 0.

[...]

PR #604, opened 2019-09-20
Keep zero values in VCF

helensch:
This is linked #602. When --sift s option is used, 0 SIFT values are not in VCF output.


at7:
I only merged it into release/98. Can you please send a PR for release/97 and postreleasefix/99?

PR #606, opened 2019-09-20
Keep zero values in VCF - 99

helensch:
This is the same as #604 on branch postreleasefix/99

Fix for when --sift s option is used, 0 SIFT values are not in VCF output.


at7:
merged into master and release/99

PR #611, opened 2019-09-30

aparton:


ens-lgil:
Merged into `master` and `release/99`

PR #614, opened 2019-10-01
Update Dockerfile

sk-sahu:
Git was missing in Stage 2, and because of that it was throwing an error.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -86,6 +86,7 @@ FROM ubuntu:18.04
 # a lot of them are required for Bio::DB::BigFile
 RUN apt-get update && apt-get -y install \
     build-essential \
+    git \
     cpanminus \
     curl \
     libmysqlclient-dev \


docker/Dockerfile
```diff
@@ -86,6 +86,7 @@ FROM ubuntu:18.04
 # a lot of them are required for Bio::DB::BigFile
 RUN apt-get update && apt-get -y install \
     build-essential \
+    git \
```
ens-lgil: Dear @sk-sahu,

Please, could you explain why you need to add `git` to the `ensembl-vep` Docker image ?

The image itself doesn't need `git` to run and we provide a new Docker image whenever ensembl-vep is updated.
Furthermore, we also try to keep the image as small as possible (minimal version), mainly for the purpose of using it as a base image in order to build custom images on top of it.

Best regards,
Laurent

sk-sahu: Hi @ens-ap5, 

Yes, I understand the reasons behind keeping it minimal. 

I was following the steps as mentioned [VEP docker documentation.](http://asia.ensembl.org/info/docs/tools/vep/script/vep_download.html#docker) 

Pulled the latest image and tried to build the cache data with following commands
```
docker pull ensemblorg/ensembl-vep
docker run -t -i ensemblorg/ensembl-vep ./vep
mkdir $HOME/vep_data
mkdir $HOME/vep_data
docker run -t -i -v $HOME/vep_data:/opt/vep/.vep ensemblorg/ensembl-vep perl INSTALL.pl
```

It thrown an error: 
```
Attempting to install Bio::DB::HTS and htslib.

>>> If this fails, try re-running with --NO_HTSLIB

  'git' command not in path. Please install git and try again.
  (or to skip Bio::DB::HTS/htslib install re-run with --NO_HTSLIB)

  On Debian/Ubuntu systems you can do this with the command:

  apt-get install git
```
So I manually build the image locally after editing `git` in the docker file and then it worked. I thought it will be helpful to suggest the same.

ens-lgil: Dear @sk-sahu,
Ah, now I see what you mean.
The documentation is actually incomplete/incaccurate. Sorry about that!
Actually the API is already installed, so you don't need to reply `y` to the question:
```
Do you want to continue installing the API (y/n)?
```
This command, in the Docker image is mainly meant to download the `VEP Caches` (data files) and/or the `VEP Plugins`.

I will update the web documentation.

Best regards,
Laurent

ens-lgil:
Updated Web documentation, regarding this Docker issue: https://github.com/Ensembl/public-plugins/pull/265

aparton:
Hi,

As we've updated our documentation to fix this issue, I'm going to close this PR now. If you have any more questions or suggestions, please feel free to reopen it or create a new one.

Kind Regards,
Andrew

Issue #610, opened 2019-09-30
Cache convertion for 98 release


Stikus:
Hello.
If I understand correctly release notes of 97 VEP release - now cache files in tabix indexed already and cache convertion step not needed anymore.
[Release notes VEP 97](https://github.com/Ensembl/ensembl-vep/releases/tag/release%2F97.0)
> tabix-indexed variant cache files are now installed by default.

Related [issue](https://github.com/Ensembl/ensembl-vep/issues/542)

But when I updated out VEP installation to 98 version (this problem was at 97 version too btw) conversion still started and too several ours both for 37 and 38 genome.

Context:
Our VEP installation command:
`perl INSTALL.pl --NO_TEST --NO_UPDATE --NO_HTSLIB --NO_BIOPERL -a ap --PLUGINSDIR "$SOFT/ensembl-vep-${VEP_VERSION}/Plugins" --PLUGINS ProteinSeqs,Downstream,Conservation,GO,G2P`

Our VEP cache install command:
`$VEPINSTALL --NO_TEST --NO_UPDATE --NO_HTSLIB --NO_BIOPERL -a cf -s homo_sapiens -y $vepCacheAssembly --CONVERT -c $VEPCACHEDIR`

If I understand perl script correctly - this [line](https://github.com/Ensembl/ensembl-vep/blob/release/98/INSTALL.pl#L1303) starts [convert_cache.pl](https://github.com/Ensembl/ensembl-vep/blob/release/98/convert_cache.pl) and this [line](https://github.com/Ensembl/ensembl-vep/blob/release/98/convert_cache.pl#L204) is check for need of convertion.
So if info.txt contains key: `var_type	tabix` - variant is skipped.

But this is our 98_GRCh37 info's before and after convertion:
Before:
```
species	homo_sapiens
assembly	GRCh37
sift	b
polyphen	b
source_polyphen	2.2.2
source_sift	sift5.2.2
source_genebuild	2011-04
source_gencode	GENCODE 19
source_assembly	GRCh37.p13
[...]

PR #635, opened 2019-11-05
Allow installer to convert cache if requested

aparton:
As described in https://github.com/Ensembl/ensembl-vep/issues/610, the CONVERT flag no longer works after changing the installer to download the indexed cache files by default

Incase there are users who want to convert old cache files, I've restored this functionality for when the --CONVERT flag is used.

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -1296,7 +1296,7 @@ sub cache() {
       closedir CACHEDIR;
     }
     
-    if(($CONVERT) && !$TEST) {
+    if(((-e $bgzip && -e $tabix) || $CONVERT) && !$TEST) {
       unless($QUIET) {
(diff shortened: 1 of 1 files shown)


INSTALL.pl
```diff
@@ -1295,6 +1299,14 @@ ()
       move("$CACHE_DIR/tmp/$species/$_", "$CACHE_DIR/$species/$_") for readdir CACHEDIR;
       closedir CACHEDIR;
     }
+    
+    if(($CONVERT) && !$TEST) {
```
ima23: Is the check for`(-e $bgzip && -e $tabix)` not needed any more? This is the line from the ticket: https://github.com/Ensembl/ensembl-vep/blob/39ee5b95677c08c58992d1f8e76a42a7f9207eda/INSTALL.pl#L1298

ima23:
merged to `release/99` and `master`

PR #639, opened 2019-11-13
Adds support for S type in cigar string

aparton:


at7:
Can you please add a test to bam_edit.t?

aparton:
Merged into release/99 and master

Issue #580, opened 2019-08-26
filter_vep requires over 100Gb of memory


fgvieira:
Dear all,

I have a VCF with SV calls annotated with VEP (`StructuralVariantOverlap.pm`), and I am trying to filter it by gnomAD AF. However, at a certain position, it uses up all available memory (around 100Gb) and is killed by the OS.

I've tried to reproduce the error on a smaller file but couldn't, so I am guessing it is some kind of memory leak. I know it is difficult to pinpoint the issue without an example, but was hoping someone might have some insight...

thanks,

dglemos:
Hi @fgvieira,

Do you know at which variant it stops running? 
I can't see an apparent reason for that to happen, as you said it is difficult to find the issue without an example. Could you send the command you used to filter the file?

fgvieira:
Yes, it stops running at:
```
2       3144802 DUP00004094     T       <DUP>   .       PASS
```
but I do not think it is this variant that is causing it, as I created a VCF with just this variant and `filter_vep` worked fine. My guess would be that, since these are SVs and some can be quite large, there is some memory that it is not being freed as it should.

```
zcat FILE.vcf.gz | filter_vep --format vcf --filter "gnomAD-SV_AF < 0.05" --only_matched
```

dglemos:
There's no apparent reason why you're having memory issues filtering variants. We're looking into it but in the meantime, it would help if you could provide your input file.

fgvieira:
Unfortunately I cannot send the file (no patient consent), but I'll see if I can replicate it with a publicly available dataset.

dglemos:
Did you replicate the issue with a different dataset?

fgvieira:
Not yet.. I've been busy quite with lately with other projects and not sure when I'll be able to look into it... maybe we can just close it for now and I'll re-open it if necessary.

dglemos:
Sorry that I couldn’t give you a more helpful response. 
I'll close the issue for now. If you have the same problem in the future feel free to re-open this issue. 

Best regards,
Diana

fgvieira:
Dear all,

I'm re-opening this issue since I just saw this behavior again with VEP 98.3 when running `filter_vep`:
`filter_vep --input_file input_cnv_germline_annot.vcf.gz --gz --filter "gnomAD_SV_AF < 0.05 or not gnomAD_SV_AF" --only_matched`

It starts with a memory usage of ~3 Gb, but it keeps increasing until ~80Gb:
[...]

PR #645, opened 2019-11-22
Improve filter_vep memory usage

dglemos:
#580 (ENSVAR-2602)
INFO and CSQ fields were being loaded for each feature overlaping a variant.

### DIFF
--- a/filter_vep
+++ b/filter_vep
@@ -235,6 +235,8 @@ sub main {
         # get info fields
         foreach my $info_field(split /\;/, (split /\s+/, $line)[7]) {
           my ($field, $value) = split /\=/, $info_field;
+          # CSQ field is not used to filter
+          next if $field eq 'CSQ';
           $main_data->{$field} = $value;
         }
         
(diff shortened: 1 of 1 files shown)


filter_vep
```diff
@@ -340,7 +342,10 @@ sub parse_line {
   my @split = split($delimiter, $line, -1);
   
   $data{$headers->[$_]} = $split[$_] for 0..$#split;
-  
+
+  # INFO is not used to filter
```
at7: This comment is misleading. INFO is used for filtering but it is done separately after reading the "main" data. Maybe a comment saying that parse_line extracts the main_data first which doesn't include INFO.

filter_vep
```diff
@@ -235,6 +235,8 @@ sub main {
         # get info fields
         foreach my $info_field(split /\;/, (split /\s+/, $line)[7]) {
           my ($field, $value) = split /\=/, $info_field;
+          # CSQ field is not used to filter
```
at7: The comment is misleading. CSQ is used for filtering but parsed later in the script.

at7:
merged into master and release/99

PR #650, opened 2019-11-28
filter_vep write header lines when input is empty

dglemos:


filter_vep
```diff
@@ -289,7 +289,13 @@ sub main {
       last if $count >= $config->{limit} + $config->{start} - 1;
     }
   }
-  
+
+  # Write header lines for empty files
+  if (!$line_number && !defined($config->{count}) && !defined($config->{list})) {
```
ima23: Looks good! I've tested it and it works for empty files. Could you please add a comment of why `!$line_number` is not a sufficient if check? Thanks.

ima23:
merged to `release/99` and `master`

PR #679, opened 2020-01-23
Filter vep - allow filtering multiple values

dglemos:
Currently, filter_vep doesn't filter values '0.7&0.5' or '0.7,0.5'. 
This fix splits the values and checks them individually.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/FilterSet.pm
+++ b/modules/Bio/EnsEMBL/VEP/FilterSet.pm
@@ -482,7 +482,9 @@ sub evaluate {
       $return = 0;
     }
     else {
-      if($input =~ /,/) {
+
+      # Filter multiple values
+      if(defined($input) && $input =~ /[0-9],/) {
         my $f = 0;
         my @input_splited = split /,/, $input;
         foreach my $value_i (@input_splited) {
--- a/modules/Bio/EnsEMBL/VEP/FilterSet.pm
+++ b/modules/Bio/EnsEMBL/VEP/FilterSet.pm
@@ -485,23 +485,16 @@ sub evaluate {
 
       # Filter multiple values
       if(defined($input) && $input =~ /[0-9](\,|\&)/) {
-        my $delimiter;
-        if($input =~ /\,/) {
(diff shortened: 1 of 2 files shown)


helensch:
Can a test be added to FilterSet.t

modules/Bio/EnsEMBL/VEP/FilterSet.pm
```diff
@@ -480,7 +482,32 @@ sub evaluate {
       $return = 0;
     }
     else {
-      $return = &$predicate($input, $value, $self);
+
+      # Filter multiple values
+      if(defined($input) && $input =~ /[0-9](\,|\&)/) {
+        my $delimiter;
+        if($input =~ /\,/) {
```
aparton: You can use $1 here to pull the delimiter out of the capture group rather than having to pattern match again

modules/Bio/EnsEMBL/VEP/FilterSet.pm
```diff
@@ -480,7 +482,32 @@ sub evaluate {
...
+          $delimiter = ',';
+        }
+        else {
+          $delimiter = '&';
+        }
+
+        my $f = 0;
+        my @input_splited = split /$delimiter/, $input;
```
aparton: I'm being picky here, but we should probably change the variable names. $f should be more descriptive and @input_splited would read better as @split_input

modules/Bio/EnsEMBL/VEP/FilterSet.pm
```diff
@@ -480,7 +482,32 @@ sub evaluate {
...
+        foreach my $value_i (@input_splited) {
+          my $predicted = &$predicate($value_i, $value, $self);
+          if($predicted) {
+            $f = 1;
+          }
+        }
+        $result = $f;
```
aparton: Is this not the same as $result = defined($predicted) ?

dglemos: defined($predicted) is always 1.

aparton:

Issue #473, opened 2019-05-30
Ensembl-vep costum annotation


shigej38:
I have a txt file in my hand where the diseases associated with the genes are written. 
How can I get ensembl-vep to add the information for the gene to the output file?

my text contents (head)
```
Approved Symbol	Phenotypes
CCV	Cataract 8, multiple types, Autosomal dominant
MYP14	Myopia 14
SCASI	Spinocerebellar ataxia, autosomal recessive 4, Autosomal recessive
ISG15	Immunodeficiency 38, Autosomal recessive
AGRN	Myasthenic syndrome, congenital, 8, with pre- and postsynaptic defects, Autosomal recessive
B3GALT6	Ehlers-Danlos syndrome, progeroid type, 2, Autosomal recessive;Spondyloepimetaphyseal dysplasia with joint laxity, type 1, with or without fractures, Autosomal recessive
DVL1	Robinow syndrome, autosomal dominant 2, Autosomal dominant
ATAD3A	Harel-Yoon syndrome, Autosomal recessive, Autosomal dominant
TMEM240	Spinocerebellar ataxia 21, Autosomal dominant

```

dglemos:
As you can read in our documentation [here](https://www.ensembl.org/info/docs/tools/vep/script/vep_custom.html), VEP can integrate custom annotation from standard format files into your results. These files must be indexed using tabix and their format can be BED, GFF, GTF, VCF.  
Your file is a tab file with only diseases and associated genes which is not suitable for vep custom annotation. Currently, we don't have any tool to perform the type of annotation you're asking. 

Thanks,
Diana

shigej38:
I want to add the names of the disease to my output, how can I do it with ensembl-vep. @dglemos

dglemos:
VEP doesn't perform the type of annotation you are asking.

shigej38:
I have two more questions.
How can I add that the variant is heterozygous or homozygous?

do I have different transcripts of the same gene written on the same line?

dglemos:
I don't understand what you mean with your first question. Could you be more specific and give an example? 
About your second question, depends on the output format you choose. It's all explained in the documentation [here](https://www.ensembl.org/info/docs/tools/vep/vep_formats.html#defaultout). Default VEP output writes one line for each transcript/regulator region; VCF output writes all data in the same line.

shigej38:
Dear @dglemos 
[...]

PR #684, opened 2020-01-29
Prevent looking up of NM/XMs in core db

aparton:
Giving the input `NM_001258273.1:c.1A>T,` to variant recoder was returning odd results, including `NM_001258273.1:c.-632A>T`

Variant recoder was generating a $vf object using a transcript generated by calling $ta->fetch_all_by_external_name on the core db, which was not an exact match. The change skips the core database and looks directly in otherfeatures for transcripts that look like RefSeq transcripts.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
@@ -168,6 +168,7 @@ sub create_VariationFeatures {
   my @errors;
 
   foreach my $core_group(@core_groups) {
+    next if (($hgvs =~ /NM_/ || $hgvs =~ /XM_/) && $core_group eq 'core');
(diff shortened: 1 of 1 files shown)


aparton:
Similar to https://github.com/Ensembl/ensembl-vep/pull/601

dglemos:
It seems to work. The HGVS genomic was also giving the wrong position but skipping the NM/XM fixes that.

ima23:
Was there a reason why https://github.com/Ensembl/ensembl-vep/pull/601/ did not go in?

dglemos:
It was a different problem. It was fetching ENST and NM transcripts with fetch_by_stable_id().

dglemos:
The problem was fixed here #473.

dglemos:
The problem was fixed here https://github.com/Ensembl/ensembl-variation/pull/473/files

dglemos:
Merged into `release/100` and `master`.

PR #686, opened 2020-02-06
Prevent lookup of NMs in core db

aparton:
Adding #684 to release 99


dglemos:
Shouldn't it be `release/99` instead of `postreleasefix/99`? The fix is already in master.

PR #692, opened 2020-02-17
Create a generic template for the VEP issues

at7:
I'm adding the template again. I will create another PR against postreleasefix/100


at7:
The template is only missing from release/99 branch. It is in the release/100 and master branches.

PR #700, opened 2020-02-25
Yearly copyright update

helensch:


dglemos:

PR #703, opened 2020-03-03
Task101/vep add more mf attribs

at7:
-- add binding matrix and transcription factors to output of motif_feature_consequences
-- this has been asked for by a REST user (https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2862)
-- I will also create a PR for ensembl-rest to check for those newly added keys and values in the VEP REST endpoints


modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -1885,7 +1894,7 @@ sub StructuralVariationOverlapAllele_to_output_hash {
   my $feature_type = (split '::', ref($feature))[-1];
 
   $hash->{Feature_type} = $feature_type;
-  $hash->{Feature}      = $feature_type eq 'MotifFeature' ? $feature->get_BindingMatrix->name : $feature->stable_id;
```
aparton: I'm sure there's a reason for it, but why has this section been removed?

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -1730,11 +1730,20 @@ sub MotifFeatureVariationAllele_to_output_hash {
...
-  $hash->{MOTIF_NAME}   = $matrix;
+  $hash->{Feature}      = $mf_stable_id;
+  $hash->{MOTIF_NAME}   = $mf_stable_id;
+  $hash->{BINDING_MATRIX} = $matrix_id;
+  my @transcription_factors = ();
+  my $associated_transcription_factor_complexes = $matrix->{associated_transcription_factor_complexes};
+  foreach my $tfc (@{$associated_transcription_factor_complexes}) {
+    push @transcription_factors, $tfc->{display_name};
+  }
+  $hash->{TRANSCRIPTION_FACTORS} = \@transcription_factors;
```
aparton: If you're adding new values to VEP output, then you'll need to:

- add a description of the field to %FIELD_DESCRIPTIONS in Constants.pm

- add a mapping between the flag and the field name in @FLAG_FIELDS in Constants.pm 

- add the field to the vep_formats.html documentation page (although you may have already done this)

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -1730,11 +1730,20 @@ sub MotifFeatureVariationAllele_to_output_hash {
...
 
+  my $mf_stable_id = $mf->stable_id;
   $hash->{Feature_type} = 'MotifFeature';
-  $hash->{Feature}      = $mf->stable_id;
-  $hash->{MOTIF_NAME}   = $matrix;
+  $hash->{Feature}      = $mf_stable_id;
+  $hash->{MOTIF_NAME}   = $mf_stable_id;
+  $hash->{BINDING_MATRIX} = $matrix_id;
+  my @transcription_factors = ();
+  my $associated_transcription_factor_complexes = $matrix->{associated_transcription_factor_complexes};
```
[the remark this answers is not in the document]
at7: I tried using the get method first. But it fails. Something is happening to the MotifFeature object when it comes back from a cache file that makes using the get method throw an error.

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
at7: This could go back to when MotifFeatures didn't have a stable id. But it is definitely wrong to return the binding matrix stable id because we are reporting the overlap with the MotifFeature.

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
at7: Thank you for the reminder. reading the definition of motif_name again made me realise that this is the binding matrix name/stable id.

aparton:
Thanks for the changes. Final request, you'll need to add a link between the '--regulatory' command line flag and the new output fields in line 127 of Constants.pm otherwise the header lines wont be correctly generated with the new data.

at7:
Tests fail because test schema doesn't contain latest updates for 101. Failures are not related to the PR.

PR #704, opened 2020-03-03
Remove old 37 transcripts from VEP (and cache files)

aparton:
No test has been added as we don't have a 37 test database.

This should remove both refseq_import and refseq_human_import transcripts.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
@@ -249,8 +249,8 @@ sub get_features_by_regions_uncached {
         # in human and mouse otherfeatures DB, there may be duplicate genes
         # skip those from analysis refseq_human_import and refseq_mouse_import
         next if $self->{core_type} eq 'otherfeatures' && $self->assembly !~ /GRCh37/i && $tr->analysis && $tr->analysis->logic_name =~ /^refseq_[a-z]+_import$/;
-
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm
```diff
@@ -249,8 +249,8 @@ sub get_features_by_regions_uncached {
         # in human and mouse otherfeatures DB, there may be duplicate genes
         # skip those from analysis refseq_human_import and refseq_mouse_import
         next if $self->{core_type} eq 'otherfeatures' && $self->assembly !~ /GRCh37/i && $tr->analysis && $tr->analysis->logic_name =~ /^refseq_[a-z]+_import$/;
-
-        if($self->{core_type} eq 'otherfeatures' && defined($tr->display_xref)){
+	next if $self->{core_type} eq 'otherfeatures' && $self->assembly =~ /GRCh37/i && $tr->analysis && $tr->analysis->logic_name =~ /^refseq.+import$/;
```
at7: As you have already said  a comment on why all those checks are needed will be helpful

aparton:
Due to the inclusion of the new RefSeq transcript set (mapped from 38) into the 37 otherfeatures database, we decided to remove all of the older, lower quality transcripts from the cache files. To do this, we filtered out all transcripts with the analysis logic_name 'refseq_import' or 'refseq_human_import'. All new transcripts have the logic name 'refseq_import_grch38'

at7:
Can you please add the comment into the Transcript.pm file? Thanks

at7:
merged into master and release/100

PR #706, opened 2020-03-04
Update dump pipeline to remove deprecated regulation methods

aparton:


aparton:
merged into release/100 and master

PR #710, opened 2020-03-12
Ontology updates for 100

aparton:


at7:
merged into master and release/100

PR #722, opened 2020-03-30
Fix warnings in VEP test scripts

helensch:
Fix warning messages in test scripts see: ENSVAR-3121


dglemos:
Merged into release/101 and master.

PR #723, opened 2020-03-30
Remove tmp files from VEP tests

helensch:
Removes files created during the run of VEP tests
See ENSVAR-2137

The FASTA file changed from test.fa to test.fa.gz as that is file in repository.
It is unpacked to test.fa as part of running the tests.


helensch:
Closing to review failing test on Runner.t

PR #729, opened 2020-04-06
Turn on shift_3prime when shift_genomic is switched on 

aparton:
Currently it requires both flags, which is unnecessary.

I'll submit a PR to postreleasefix/101 when this is approved

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Runner.pm
+++ b/modules/Bio/EnsEMBL/VEP/Runner.pm
@@ -766,6 +766,13 @@ sub post_setup_checks {
     $self->status_msg("INFO: --shift_hgvs has been set to zero, setting shift_3prime to 0\n");
     $self->param('shift_3prime', 0);
   }
+  
+  if(defined($self->param('shift_genomic')) && $self->param('shift_genomic') eq 1 ) {
+    if(defined($self->param('shift_3prime')) && $self->param('shift_3prime') eq 0 ) {
+      $self->status_msg("INFO: --shift_genomic has been set to 1, setting shift_3prime to 1\n");
+    }
+    $self->param('shift_3prime', 1);
+  }
 
   return 1;
 }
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/VEP/Runner.pm
```diff
@@ -766,6 +766,13 @@ sub post_setup_checks {
     $self->status_msg("INFO: --shift_hgvs has been set to zero, setting shift_3prime to 0\n");
     $self->param('shift_3prime', 0);
   }
+  
+  if(defined($self->param('shift_genomic')) && $self->param('shift_genomic') eq 1 ) {
+    if(defined($self->param('shift_3prime')) && $self->param('shift_3prime') eq 0 ) {
```
dglemos, line 771: The message could also include when shift_3prime is not defined. In that case users would know setting shift_genomic=1 will automatically set shift_3prime=1 too.

modules/Bio/EnsEMBL/VEP/Runner.pm
```diff
@@ -766,6 +766,13 @@ sub post_setup_checks {
     $self->status_msg("INFO: --shift_hgvs has been set to zero, setting shift_3prime to 0\n");
     $self->param('shift_3prime', 0);
   }
+  
+  if(defined($self->param('shift_genomic')) && $self->param('shift_genomic') eq 1 ) {
```
helensch, line 770: As numerical comparison suggest using "==" or  just ($self->param('shift_genomic'))

modules/Bio/EnsEMBL/VEP/Runner.pm
aparton: I don't want to add this change - shift_3prime is being defined by default to 0/1 in the vep module, so any occasion where it's undefined would be in our testing modules and I don't want to print this warning at that point

PR #735, opened 2020-04-16
Added shift length flag

aparton:
I'll submit another PR to 101 when this is approved


dglemos:
Thanks for changing the description. After running a few tests everything looks fine.

PR #742, opened 2020-04-24
Shift length updates for 101

aparton:


dglemos:

PR #752, opened 2020-05-11
SPDI reference sequence

dglemos:
Not necessary to check if the reference sequence is not genomic.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser/SPDI.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser/SPDI.pm
@@ -184,11 +184,7 @@ sub create_VariationFeatures {
     return $self->create_VariationFeatures;
   }
 
-  # throws error if reference sequence is not genomic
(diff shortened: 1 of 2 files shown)


dglemos:
One test is failing because it needs code from https://github.com/Ensembl/ensembl-variation/pull/592

modules/Bio/EnsEMBL/VEP/Parser/SPDI.pm
```diff
@@ -184,11 +184,7 @@ sub create_VariationFeatures {
     return $self->create_VariationFeatures;
   }
 
-  # throws error if reference sequence is not genomic
-  if($spdi !~ /^(NC_|CHR|([0-9]{1,2}|X|Y|MT):)/i) {
-    my $ref_seq = (split(':', $spdi))[0];
-    throw("Invalid reference sequence '$ref_seq' as SPDI reference");
-  }
+  my $ref_seq = (split(':', $spdi))[0];
```
sarahhunt, line 187: This isn't needed.

PR #756, opened 2020-05-14
added Arabidopsis thaliana VCF sample

brunocontrerasmoreira:
Hi, I have been testing the VEP script on plants and found out there was no suitable sample VCF file.
I am adding one here from https://europepmc.org/article/MED/27293186 for Arabidopsis thaliana


examples/arabidopsis_thaliana.TAIR10.vcf
```diff
@@ -0,0 +1,25 @@
+##fileformat=VCFv4.1
+# source: intersection_10001.vcf.gz
+# https://europepmc.org/article/MED/27293186
+1	135	.	C	T	28	PASS	DP=5	GT:GQ:DP	1|1:28:5
```
dglemos, line 4: Could you add the header line names? Something like: 
`#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	SAMPLE`

examples/arabidopsis_thaliana.TAIR10.vcf
```diff
@@ -0,0 +1,25 @@
+##fileformat=VCFv4.1
+# source: intersection_10001.vcf.gz
```
dglemos, line 2: Thanks for adding a new file.
Here should be two '##' without space between 'source'. The same in L3.

brunocontrerasmoreira: I have fixed it following your suggestions, thanks!

brunocontrerasmoreira:
I have done as requested @dlemos, obrigado

dglemos:
Thanks for the changes!

PR #758, opened 2020-05-15
Bugfix for --json and --vcf with --shift_3prime

aparton:
101 version of #755


at7:
merged into master and relase/101

Issue #759, opened 2020-05-17
ClinVar fields not consistent 


EugeneEA:
## Describe the issue
I'm trying to annotate a toy example of two SNP and get CLIN_SIG and CLNREVSTAT fields. 
here is the input vcf wo header:
`#CHROM	POS	ID	REF	ALT	QUAL	FILTER
chr7	140749365 	.	A	T	50	PASS
chr7	140753339	.	G	C	50	PASS
`

both SNP are in BRAF protein, Pathogenic and have a clinVar records (I'm using latest version of clinvar as custom annotation, and check it there). 

But the problem here is - for some reason there are no CLIN_SIG and CLNREVSTAT fields for the first SNP. I've double check the raw clinVar file and it is there, so I'm quite confused...

###  Here is the Full VEP command line (run from the docker container)

`~/src/ensembl-vep$ time ./vep --cache --offline --format vcf --vcf --force_overwrite --dir_cache /opt/vep/.vep/ --input_file /opt/vep/.vep/input/test_chr_renamed.vcf --output_file /opt/vep/.vep/output/my_output.vcf --buffer_size 10000  --fork 30 --custom /opt/vep/.vep/custom/clinvar.vcf.gz,ClinVar,vcf,exact,0,CLNSIG,CLNREVSTAT,CLNDN`

and here is the VEP output (sorry for the mess):

`#CHROM	POS	ID	REF	ALT	QUAL	FILTER
[...]

PR #762, opened 2020-05-18
Trim whitespace from input coordinates

aparton:
As reported in #759, the VCF parser is not matching colocated variants when there is unexpected whitespace found in the coordinate field. Trimming whitespace from here should fix the issue


aparton:
Not adding, breaks VCF spec

PR #767, opened 2020-05-21
Set port to 3337 when assembly is set to grch37

aparton:
No test has been added as it would require a grch37 test database to be set up


sarahhunt:
We could lose the "If using human GRCh37 add --port 3337" message from:

- modules/Bio/EnsEMBL/VEP/BaseRunner.pm
- modules/Bio/EnsEMBL/VEP/CacheDir.pm

now

PR #769, opened 2020-05-21
Check so_term if so_name is empty

aparton:
The feature_type value for regulatory features was missing - caused by changes in the funcgen api for 100.


sarahhunt:
@aparton - can you add a check for this to e!101?

PR #787, opened 2020-06-23
Example Update

aparton:
As reported by Michal on ensembl-variation on 23/06/2020 - the example data for GRCh38 contains an error - alt/ref are the wrong way around: http://www.ensembl.org/Homo_sapiens/Variation/Explore?r=21:37072063-37073063;v=rs2507733;vdb=variation;vf=51868053


at7:

PR #799, opened 2020-07-09
Bugfix101/vcf minimal rejoin

at7:
Rejoin alleles after minimal was applied. This is done in all other output factories but was missing from the VCF output factory. This is a bug reported bu a user on release/100.
I will also create a PR for 102 and master and ideally for 100 too.


at7:
Thank you for the review. Let me do the release/100 and postreleasefix/102 PRs first before you can merge them all in one go. Thank you

PR #803, opened 2020-07-13
Bugfix102/vcf minimal rejoin

at7:
Same as #799


aparton:
Merged to release/102 and master

Issue #690, opened 2020-02-12
filter_vep add label to FILTER field


fgvieira:
Dear all, 

as it is now, all variants that match the filtering criteria are excluded.
Would it be possible to add an option to add (or replace) a label under the FILTER field? Like `bcftools filter` option `--soft-filter --mode +`. For example, the variant:
`1       17556   .       C       T       64.6    PASS`
would become:
`1       17556   .       C       T       64.6    failed_af`

or, if variant was:
`1       17556   .       C       T       64.6    failed_QC`
it would become:
`1       17556   .       C       T       64.6    failed_QC,failed_af`

thanks,

dglemos:
Hi @fgvieira,

Just to know if I understand what your suggestion: you mean to change the value in 'FILTER' column (in vcf format) to a value that says this variant passed the `filter_vep` criteria?

fgvieira:
well, change (or append) the FILTER column to some specific label if it did not match the `filter_vep` criteria.

dglemos:
Thank you for the suggestion. We're going to look into adding this option to filter_vep.

dglemos:
Hi @fgvieira,
Recently we added a new option to filter_vep to flag variants failing the filter instead of deleting them from the final file. This new option adds a flag to the FILTER value in the VCF file. This flag does not specify which parameters failed, it only indicates if the variant fails or not: `filter_vep_pass` means the variant passed the filter; `filter_vep_fail` variant failed the filter. 
I hope this is helpful, I'm going to close this issue but feel free to open a new one if you have more suggestions or questions. 

Best wishes,
Diana

dglemos:
The new Ensembl release is out. For this release, we added the option `soft_filter` to filter_vep that doesn't exclude the failing variants from the output. Instead, it adds a flag to the FILTER value in the VCF. 
I'm going to close this issue now but if you have related questions feel free to reopen it. 
Best wishes,
Diana

PR #809, opened 2020-07-22
Add option to filter_vep

dglemos:
User request #690 to have an option to flag which variants pass or fail the filter instead of deleting them.

### DIFF
--- a/filter_vep
+++ b/filter_vep
@@ -85,6 +85,7 @@ sub configure {
     'only_matched',            # rewrite CSQ field in VCF with only matched "blobs"
     'vcf_info_field=s',        # VCF INFO field name to parse
     'soft_filter',             # do not exclude variants failing the filter - add a flag to the FILTER value in the vcf
+                               # 'vep_filter_pass' variant passed the filter; 'vep_filter_fail' variant failed the filter 
 
     'ontology|y',              # use ontology for matching consequence terms
     'host=s',                  # DB options
@@ -297,7 +298,7 @@ sub main {
         my @split_line = split /\t/, $line;
         my $option;
 
-        $option = $line_pass ? "PASS" : "FAILED";
+        $option = $line_pass ? "vep_filter_pass" : "vep_filter_fail";
         if($split_line[6] ne '.') {
           $option = join(",", $split_line[6], $option);
         }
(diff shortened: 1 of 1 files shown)


filter_vep
```diff
@@ -84,7 +84,8 @@ sub configure {
     'gz',                      # force read as gzipped
     'only_matched',            # rewrite CSQ field in VCF with only matched "blobs"
     'vcf_info_field=s',        # VCF INFO field name to parse
-    
+    'soft_filter',             # do not exclude failed variants, only changes the FILTER value in the vcf
```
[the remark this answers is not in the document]
dglemos: "variants failing filter" sounds correct. I'll add a check - if FILTER has values then it appends the flag instead of replacing it.

dglemos:
Thanks for noticing the bug. I also noticed that soft_filter doesn't work all the time even when some matches.

dglemos:
@helensch I made all the changes. I also tested the vcf as suggested and it passes the validator.

filter_vep
```diff
@@ -196,13 +196,20 @@ sub main {
       # parse headers before processing input
       if(!(scalar @headers)) {
         throw("ERROR: No headers found in input file\n") unless scalar @raw_headers;
-        
+
+        if($config->{soft_filter} && (scalar @raw_headers)) {
+          my $chr_header = pop @raw_headers;
+          push @raw_headers, "##FILTER=<ID=vep_filter_pass,Description=\"Variant pass vep filter\">";
```
helensch, line 202: Suggest maybe using filter_vep instead of vep filter
Suggest just saying "passes filter_vep" "fails filter_vep".
Examples from VCF spec just have criteria.

filter_vep
```diff
@@ -194,13 +196,20 @@ sub main {
       # parse headers before processing input
       if(!(scalar @headers)) {
         throw("ERROR: No headers found in input file\n") unless scalar @raw_headers;
-        
+
+        if($config->{soft_filter} && (scalar @raw_headers)) {
+          my $chr_header = pop @raw_headers;
+          push @raw_headers, "##FILTER=<ID=passes filter_vep,Description=\"Variant pass vep filter\">";
```
helensch, line 202: Sorry,  I was suggesting a change to the Description values not the ID value.
From the VCF spec, the ID value  used in filter should be "String, no white-space or semi-colons permitted"
I recommend the IDs be filter_vep_pass and filter_vep_fail

helensch:

Issue #806, opened 2020-07-20
Using both --minimal and --fork options with plugins causes VEP to die


david-a-parry:
## Describe the issue
Using both `--minimal` and `--fork` options with plugins causes VEP to die:

```
Can't call method "feature" on an undefined value at <path removed>/ensembl-vep/Bio/EnsEMBL/Variation/BaseVariationFeatureOverlapAllele.pm line 196.
Died in forked process 50898
```
This is not specific to any particular plugin; I have tried running with several plugins including LoFtool, SpliceRegion and TSSDistance and the same issue arises.

The issue only occurs when using BOTH `--minimal` and e.g. `--fork 4`. I do not get the error when running with `--minimal` flag without forks or when running with multiple forks without the `--minimal` flag.

From the limited debugging I've attempted, the error appears to occur at intergenic variants.

## Additional information
Please fill in the following sections to help us find the source of your issue as quickly as possible.

### System
- VEP version: 100.3
- VEP Cache version: 100
- Perl version: 5.30
- OS: Ubuntu 20.04
- tabix installed: yes

### Full VEP command line
```
./vep --offline -i fork_error_examples.vcf.gz --dir_plugins VEP_plugins/ --plugin LoFtool --fork 4 --minimal
```

### Full error message
> Can't call method "feature" on an undefined value at [path removed]/ensembl-vep/Bio/EnsEMBL/Variation/BaseVariationFeatureOverlapAllele.pm line 196.
Died in forked process 50898

### Data files (if applicable)
- The input file
[fork_error_examples.vcf.gz](https://github.com/Ensembl/ensembl-vep/files/4946763/fork_error_examples.vcf.gz)

dglemos:
Hello David,
thank you for reporting this error. We are looking into it and we'll let you know when we know what is causing the sudden stop.
Best wishes,
Diana

at7:
Dear David,

this one is a more complex bug where the rejoining of intergenic variants with more than one alternative allele is causing a problem. We have written a fix for the bug which still needs to be reviewed. Because of the slightly more complex nature of the bug this might take some time. But we will keep you updated about any progress and then distribute the fix as soon as possible.

Best wishes,
Anja

at7:
Dear David,

this has been fixed now for release/100. Could you please pull the latest changes from the release/100 branch? Thank you for providing your vep command and input data. This really helped us to reproduce the error and work on a fix. Please let us know if you continue seeing any problems.

Best wishes,
[...]

PR #811, opened 2020-07-27
Bugfix102/minimal intergenic

at7:
Rejoining intergenic variant alleles after minimising them caused an error for variants with more than one alt allele. The bug is reported in this issue #806
After rejoining the intergenic variant alleles the base_variation_feature_overlap object was missing for the allele that was merged into the original. Adding the base_variation_feature_overlap from the original is fixing the problem.


at7:

PR #822, opened 2020-08-04
Add UniProt isoforms 

aparton:
Includes uniprot isoforms within VEP output by adding a new output field when --uniprot is specified. Also included in VEP cache


sarahhunt:
Thanks @aparton - looks good. 
Does the expected output in the test need a another '-' for the UniProt isoform column?

PR #823, opened 2020-08-06
include mouse strains

at7:
At the moment strains and breeds are not shown in the list of species on VEP tools. We already dump dog and pig breeds and we want to show the new sheep reference and the previous sheep reference which has a widely used variation database.
In order to make it easier for the web code to show all strains and breeds and not make exceptions for mouse we will also dump the mouse strains from release 102.


at7:

Issue #833, opened 2020-08-20
Feature request: VCF header should contain gene GTF version


davmlaw:
I need to know the gene annotation release version used, but this is not possible to automate at the moment.

**Ensembl**

You know the Ensembl release (eg 100)

GRCh38 is being updated, so you know it's release 100.
GRCh37 is not being updated, so going to ftp://ftp.ensembl.org/pub/grch37/release-100/gtf/homo_sapiens/Homo_sapiens.GRCh37.87.gtf.gz - you find out it's v87

**RefSeq**

For VEP release 100:

GRCh37 - refseq column contains "01_2015" - however if you go to the [VEP cache page](https://m.ensembl.org/info/docs/tools/vep/script/vep_cache.html) you can see that it lists the actual GTF for GRCh37 as GCF_000001405.25_GRCh37.p13_genomic.gff 
GRCh38 - refseq column contains "2019-12-11 22:24:10 - GCF_000001405.39_GRCh38.p13_genomic.gff" so you can use this to work out the version

### Suggested fix

Add a new entry to VEP header (eg "genes-gtf") which consistently describes the GTF used in that VEP annotation version

dglemos:
Hi @davmlaw,
Thank you for the explanation and the suggestion. We will look into it and will let you know of any changes made. 
Best wishes,
Diana

dglemos:
We have updated the VEP cache RefSeq info, it now lists `2019-10-24 23:10:14 - GCF_000001405.25_GRCh37.p13_genomic.gff` instead of `01_2015`. The date `2019-10-24` indicates when the RefSeq file was imported into Ensembl. 
I'm going to close the ticket but if you have more questions feel free to reopen it.
Best wishes,
Diana

davmlaw:
Hi, I am unable to re-open this issue.

Thanks for identifying the RefSeq GFF

Unfortunately there is still no way to see that the Ensembl GRC37 GFF is still in effect v87 - it always lists it as the current Ensembl release, even though this is not true, ie:

http://ftp.ensembl.org/pub/grch37/release-108/gff3/homo_sapiens/ has the 87 GFF

davmlaw:
Also - the date RefSeq release date was 2020-10-22. The date it was loaded into Ensembl - ie 2020-10-26 17:03:42 is close to this but you have to do some work to figure it out.

Ideally - you'd include the URL of where to get the GTF - that would work for both RefSeq and ensembl, ie GRCh38 108 could just link to 

[...]

PR #834, opened 2020-08-21
RefSeq GRCh37 cache 

dglemos:
Display the correct db_version in cache files 
Related to issue #833


aparton:
merged into release/102 and master

PR #839, opened 2020-08-28
Change Variant Recoder output structure

dglemos:
Variant Recoder output to be specific by allele. 
COSMIC and HGMD ids are not linked to any allele - in the output, these ids are displayed for all alleles. 
Example to test: rs1160446270

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/VariantRecoder.pm
+++ b/modules/Bio/EnsEMBL/VEP/VariantRecoder.pm
@@ -70,8 +70,6 @@ use Bio::EnsEMBL::Variation::VariationFeature;
 use Bio::EnsEMBL::Variation::Utils::VEP;
 use Bio::EnsEMBL::Utils::Sequence qw(reverse_comp);
 
-use Data::Dumper;
-
 =head2 new
 
   Arg 1      : hashref $config
--- a/modules/Bio/EnsEMBL/VEP/VariantRecoder.pm
+++ b/modules/Bio/EnsEMBL/VEP/VariantRecoder.pm
@@ -67,6 +67,7 @@ use Bio::EnsEMBL::Utils::Exception qw(throw warning);
 use Bio::EnsEMBL::VEP::Runner;
 use Bio::EnsEMBL::VEP::Utils qw(find_in_ref merge_arrays);
 
+use Data::Dumper;
 
 =head2 new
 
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/VEP/VariantRecoder.pm
```diff
@@ -245,12 +247,93 @@ sub _get_all_results {
...
+
+    $line_by_allele{'consequences'} = \%allele_consequence;
+
+    # Parse vcf string and build a hash by allele
+    my %vcf_string_by_allele;
+    if($keys_no_allele{'vcf_string'}) {
+      # Minimise alleles
+      my $minimise_alleles_hash = $self->_minimise_allele($line->{'allele_string'}, $line->{start}, $line->{end}, $line->{strand});
+      # If there is more than one vcf_string then it's an array
+      if(ref($line->{'vcf_string'})) {
```
at7, line 284: Do you have an example when there is more than one vcf_string? Don't you need to compare ref($line->{'vcf_string'})) against 'ARRAY'?

modules/Bio/EnsEMBL/VEP/VariantRecoder.pm
```diff
@@ -245,12 +247,93 @@ sub _get_all_results {
...
+        if($co_var->{'allele_string'} =~ /COSMIC|HGMD/) {
+          push @ids_no_allele, $co_var->{'id'}; 
+        }
+        else {
+          my @split_allele = split /\//, $co_var->{'allele_string'};
+          # delete ref allele
+          shift @split_allele;
+          foreach my $allele (@split_allele) {
+            my $allele_rev = $allele;
+            reverse_comp(\$allele_rev);
```
at7, line 311: I would only reverse if $allele_consequence{$allele} doesn't exist. Then you save a tiny bit of time

at7 (approved):
Overall I think it is very clever how you create a separate result for each allele. I also think it is very readable code but it is still not easy to fully understand the steps. I would suggest that you add a lot more comments explaining all the minimising and matching steps.

t/VariantRecoder.t
```diff
@@ -125,38 +165,45 @@ SKIP: {
...
+# Tests need to be allele specific
+#  $vr->param('input_file', $test_cfg->{vr_vcf});
+#  my $results = $vr->recode_all();
 
-  is_deeply(
-    [map {$_->{hgvsg}->[0]} @$results],
-    ['21:g.25585733C>T', '21:g.25587701T>C', '21:g.25587758G>A'],
-    'recode_all'
-  );
+#  is_deeply(
```
at7, line 185: This hashed out code needs to be deleted otherwise it causes confusion.

at7:
I'm happy to merge the PR before we branch for release/103. I created a jira ticket for you as a reminder to update the merged code.

dglemos:
@at7 thanks for the review! I will address all comments in a new PR.

modules/Bio/EnsEMBL/VEP/VariantRecoder.pm
dglemos: I added an example with more than one vcf_string to the comments. I don't compare ref($line->{'vcf_string'})) against 'ARRAY' because here I just want to know if it's a string or not.

PR #856, opened 2020-09-22
Update variant recoder by allele 

dglemos:
ENSVAR-3494


dglemos:
@at7 I added a few comments and examples. Let me know if I need to add more.

dglemos:
Input "ENST00000346798.8:p.Glu693Gln" the vcf_string was missing from the output because the alleles were 'C/G' but the allele_string is 'G/C', the comparison was failing.

at7:
merged into master and release/103

PR #869, opened 2020-10-20
Variant recoder output bug fix

dglemos:
Some allele strings didn't match the vcf string alleles and the vcf_string was missing from the output. 
Example: rs774003114


at7 (approved):

at7:
merged into master and release/103

PR #872, opened 2020-10-22
Fix missing ref_allele when using fields -e102

ima23:
User reported issue `--fields REF_ALLELE` does not work, column ends up having `-`.
Same issue for `SYMBOL` fixed too.

jira ticket: ENSVAR-3575

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
@@ -1249,7 +1249,8 @@ sub VariationFeatureOverlapAllele_to_output_hash {
   $hash->{ALLELE_NUM} = $vfoa->allele_number if $self->{allele_number};
 
   # reference allele
-  $hash->{REF_ALLELE} = $vf->ref_allele_string if $self->{show_ref_allele};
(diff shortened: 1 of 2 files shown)


aparton (commented):
While this fix is a solution to the initial user issue, using the flag '--show_ref_allele' would also add the ref allele to the output while `--fields` is used, without the need for a code change.

Additionally, this change would allow for the REF_ALLELE and SYMBOL fields to be shown if selected in --fields while their corresponding flags are not switched on, similar other flags (e.g. canonical or frequency info) would not follow the same behaviour. I don't believe we want to support this, as some of these flags do work before we reach the output factory, so we could end up with inconsistent behaviour.

ima23:
@aparton thanks for checking the PR. I agree the `--show_ref_allele` is a work-around to the issue, however the original issue should be fixed.

There is no mention in the docs that the `--fields` should work only when the other flags are on, or have I missed anything? Can we have a chat on this, would be great to get the fix in for e102.

ima23:
Due to the way VEP works in the background, the fix will be to add to the documentation that `--show_ref_allele` and `--symbol` is required for `--fields 'REF_ALLELE,SYMBOL' ` to retrieve any information. Therefore this code update is not required any more.

PR #873, opened 2020-10-23
Fix missing ref_allele when using fields  -e103

ima23:
same as https://github.com/Ensembl/ensembl-vep/pull/872, now for e103


ima23:
Due to the way VEP works in the background, the fix will be to add to the documentation that `--show_ref_allele` and `--symbol` is required for `--fields 'REF_ALLELE,SYMBOL' ` to retrieve any information. Therefore this code update is not required any more.

PR #880, opened 2020-11-11
Add mane_plus_clinical

aparton:


aparton:
Current strategy is:
- There should be two flags: --mane (provides MANE_SELECT and MANE_PLUS_CLINICAL output) and --mane_select (provides only MANE_SELECT output)
- Should be accessible through both --database and --cache

ima23 (commented):
Functionality is there, tested on 102 cache and 103 db, only 103 contains mane plus. The only comment is that the test is not really a test, but a place holder. Is there time to add an actual test?

ima23:
Tests are based on cache, updating of test cache will be done at a later date ENSVAR-3753

ima23:
merged to `release/103` and `master`

Issue #875, opened 2020-10-28
Variants on scaffolds without entry in gff3 are removed from output


jd3234:
I am trying to annotate a BCF with a custom gff3 but I am getting the `... not found in annotation sources of synonyms` error.  Variants on these scaffolds are not written in the output of VEP. The "problem" with the gff3 is, that some scaffolds have no gene annotation and hence there is no entry for the scaffold in the gff3. 
A workaround I found is to add a chromosome annotation in the gff3. But this seems cumbersome... Is there a paramater I can switch on to include variants that can not be annotated? I tried to add `--dont_skip` but that did not do the trick. 
If there is no parameter would it be possible to add a feature like that for the future?

### System
- VEP version: 100.2 
- VEP Cache version: -
- Perl version: 5.26
- OS: Ubuntu, CentOS
- tabix installed ? yes

### Full VEP command line
```
bcftools view in.bcf | vep --gff customgff.gff3.gz --fasta custom.fasta --format vcf --dont_skip --vcf -o annotated.vcf
```

### Full error message
> WARNING: Chromosome SEQ_1 not found in annotation sources or synonyms on line 1

at7:
Hi,

using --dont_skip will still write variants, that for example fall on unrecognised sequences, into the output file.

Best wishes,
Anja

jd3234:
Dear Anja,

thanks for your reply. Unfortunately using --dont_skip does not help. I also tried using a gtf but no success either. I also updated to version 101 but didn't help either.

at7:
You are correct that you still get the warning message but the variant should be written into your output file. You could try to use [cache files](https://www.ensembl.org/info/docs/tools/vep/script/vep_cache.html#cache) as an additional annotation source. Could you please send me some of the scaffold names and I can check if the cache files contain them?
Thank you,
Anja

jd3234:
Hi Anja,
but unfortunately the variants are not in the output. I can not use cache files as I am working with unpublished data.
I can send a test example if you can share an email address.
Best,
Janina

at7:
Hi Janina,
it would help me to debug the problem if you could send some example data. Could you please send the files to this address helpdesk [at] ensembl.org and refer to vep-issue 875? Our outreach team will forward the files to me.
Thank you,
Anja

at7:
Hi Janina,
I received the files and will get back to you when I know more.
Best wishes,
Anja

at7:
Dear Janina,
[...]

PR #881, opened 2020-11-11
Bugfix103/dont skip

at7:
Even with the dont_skip option variants that don't overlap transcripts in the GFF file where not included in the output. $parser->next() is using exit 1 in the [ensembl-io](https://github.com/Ensembl/ensembl-io/blob/release/101/modules/Bio/EnsEMBL/IO/TabixParser.pm#L109) when the iterator is not defined which I think is the problem here. The issue was reported here https://github.com/Ensembl/ensembl-vep/issues/875. I add some test files to PANDA/anja/vep_issue_875/


bryanlaura736 (approved):

bryanlaura736 (approved):

at7:
@aparton can you please only merge this into release/104 after reviewing the PR? I will create the same PR against postreleasefix/104. Thank you

aparton:
Looks good. If you tag me in the 104 PR as well then I can merge that too

PR #917, opened 2021-01-18
Yearly copyright update

helensch:
ENSVAR-3879


aparton:

PR #885, opened 2020-11-16
Fix travis by updating VEP test with alias_to updates

aparton:

PR #929, opened 2021-02-16
Remove alias_to from test

aparton:


aparton:
Initial change was here: https://github.com/Ensembl/ensembl-vep/pull/885

PR #934, opened 2021-02-17
Variant Recoder report variant synonyms

dglemos:
ENSVAR-2692
Dependent on https://github.com/Ensembl/ensembl-variation/pull/717


modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -194,6 +194,7 @@ sub new {
     mirna
     ambiguity
     var_synonyms
+    variant_synonyms
```
aparton, line 197: Should we use the current `var_synonyms` instead of adding a new output field?

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -889,6 +890,13 @@ sub VariationFeature_to_output_hash {
     }
   }
 
+  # get variation synonyms for Variant Recoder
+  if($self->{variant_synonyms} || (defined($self->{_config}->{_params}->{fields}) && grep(/variant_synonyms/, @{$self->{_config}->{_params}->{fields}}))){
```
aparton, line 894: Switching parameters on with fields can get us into some problems and lead to duplication of data within VEP, for example:
perl vep -id rs699 --cache --dir $CACHE_DIR --var_synonyms --fields var_synonyms,variant_synonyms

aparton (changes requested):

dglemos:
@aparton I deleted the new flag `variant_synonyms`. Both Variant Recoder and VEP use `var_synonyms`.
$0 is used to check which tool is running - in the tests Variant Recoder is 't/VariantRecoder.t' while the tool is 'variant_recoder', that is why there are two checks `$0 =~ /variant_recoder/ || $0 =~ /VariantRecoder/`

dglemos:
@aparton I added flag `is_vr` to VariantRecoder.pm. Later we can use this flag to check which tool is running. 
It works on Web, here is a sandbox to test: http://ves-hx2-76.ebi.ac.uk:7040/info/docs/tools/vep/index.html

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -889,6 +889,14 @@ sub VariationFeature_to_output_hash {
     }
   }
 
+  # get variation synonyms for Variant Recoder
+  # if the tool is Variant Recoder fetches the variation synonyms from the database
+  if($self->{_config}->{_params}->{is_vr} && ($self->{var_synonyms} || (defined($self->{_config}->{_params}->{fields}) && grep(/var_synonyms/, @{$self->{_config}->{_params}->{fields}})))){
```
aparton, line 894: Do we need the `(defined($self->{_config}->{_params}->{fields}) && grep(/var_synonyms/, @{$self->{_config}->{_params}->{fields}}` section? Using --fields to switch parameters on can lead to issues if we ever require code to be run before we reach the OutputFactory

aparton (changes requested):

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
dglemos: Deleted the block of code. `fields` doesn't switch the option on anymore.

PR #947, opened 2021-02-26
Post process var synonyms to control delimiter

aparton:
Relatively small change. There is no test added currently as were aiming to get this change in before the VM is generated


dglemos (approved):

dglemos:
Tested on the sandbox. The output links look alright.

dglemos:

Issue #938, opened 2021-02-23
Issue with the allele frequency flags


fardokhtsadat:
Hello,

I am using VEP verison 103.1, and I am trying to extract allele-frequencies from the cache data (homo_sapiens_refseq_vep_102_GRCh38.tar.gz).
The allele frequencies that I am interested in are --af, --af_1kg and --af_gnomad.

Issues and questions:
1. The first issue that I faced is that the flag --af returns the allele frequencies for all the flags (gnomad, 1kg, esp) while --af_1kg and --af_gnomad are not used. While --af returns all other allele frequencies, it never returns the global AF value. 
(using only --af or  only -af_1kg and --af_gnomad return the same results.)

Why does this happen, and why no global allele frequency (AF) is returned?

2.The second issue is that, the output shows different formats. For instance, in example 1 in the colocated_variants, the gnomad allele frequenices are shown as (gnomAD_ + CAPITAL LETTERS), while in the second example the gnomad allele frequencies are shown as (gnomAD_ + small letters). 
What is the reason for this inconsistancy?

The other difference is that in the second example the allele frequnecies are a part of the field called 'frequencies', but this is not the case. 
Why does this happen?

example 1:
` 'colocated_variants': [{'start': 153935318,
   'gnomAD_NFE': 'A:0',
   'gnomAD_ASJ': 'A:0',
   'gnomAD': 'A:0',
   'gnomAD_FIN': 'A:0',
   'gnomAD_OTH': 'A:0',
   'end': 153935318,
   'gnomAD_AFR': 'A:0',
   'allele_string': 'C/A/G/T',
   'gnomAD_EAS': 'A:0',
   'seq_region_name': 'X',
   'gnomAD_AMR': 'A:0',
   'strand': 1,
   'gnomAD_SAS': 'A:0',
   'id': 'rs868959703'}]`
   
example 2:

` 'colocated_variants': [{'somatic': 1,
  'id': 'COSV66935033',
  'strand': 1,
  'seq_region_name': '11',
  'start': 18137446,
  'allele_string': 'COSMIC_MUTATION',
  'phenotype_or_disease': 1,
  'end': 18137446,
  'var_synonyms': {'COSMIC': ['COSM7248823']}},
 {'end': 18137446,
  'minor_allele': 'C',
  'allele_string': 'T/C',
  'seq_region_name': '11',
  'strand': 1,
  'id': 'rs12291017',
  'var_synonyms': {'UniProt': ['VAR_025507']},
[...]

PR #950, opened 2021-03-03
VEP JSON output - remove frequencies of non-matching co-located alleles

helensch:
This is related to  #938. For further details and examples see: ENSVAR-4006

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm
@@ -124,6 +124,13 @@ my @LIST_FIELDS = qw(
   var_synonyms
 );
 
+my @FREQ_FIELDS = qw(
+  AFR AMR ASN EAS SAS EUR
+  AA EA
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm
```diff
@@ -124,6 +124,13 @@ my @LIST_FIELDS = qw(
   var_synonyms
 );
 
+my @FREQ_FIELDS = qw(
```
at7, line 127: Could you please use the new array also in this loop here https://github.com/Ensembl/ensembl-vep/blob/release/103/modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm#L396

at7 (approved):
This fixes an oversight when working with multi-allelic variants where only a subset of the alleles is provided by the user. It is absolutely correct to remove all remaining frequency keys from the existing hash.

at7:
I realised that I already approved the PR. Sorry for the confusion. I will approve it again after you had a chance to update the array.

helensch:
@at7  thanks for the review. I have modified to use @FREQ_FIELDS when setting frequencies

at7 (approved):

at7:
merged into master and release/104

PR #960, opened 2021-03-29
use r2.1.1 for consistency

at7:
We used r2.1 before and it is better to use r2.1.1 for consistency.


aparton:

PR #963, opened 2021-04-01
Task104/vep dumps update gnomad 3

at7:
The new gnomAD files use chr prefixes which we need to consider in the dumps.


aparton:
merged into release/104 and master

PR #965, opened 2021-04-06
Update README 

dglemos:
Update Variant Recoder output example


aparton:

Issue #966, opened 2021-04-07
Updated install:  no cache found for homo_sapiens


alanhoyle:
## Describe the issue
I just installed a copy of VEP 103 through Bioconda and installed the homo_sapiens GRCh38 with 103 as well (I have an existing VEP 97 install too).  When I try to run VEP using the `--cache`, it dies quickly, saying `MSG: ERROR: No cache found for homo_sapiens, version 103`, however `INSTALL.p`l (aka `vep_install` to conda)  finds that it the cache exists and won't install a new one.  

I've also tried adding --offline and that makes no difference

## Additional information
Please fill in the following sections to help us find the source of your issue as quickly as possible.

### System
- VEP version: 103
- VEP Cache version: 103
- Perl version: 5.26
- OS: CentOS 7
- tabix installed YES

### Full VEP command line
Both the install command and the actual vep call are below:  
```
vep_install -a cf -c ${VEP_CACHE} -s homo_sapiens --assembly GRCh38 --CONVERT
vep --cache ${VEP_CACHE} -i somatic.vcf.gz -a GRCh38 -o ~/tmp/somatic.vep.vcf.gz
```

### Full error message
```
 $ vep --cache ${VEP_CACHE} -i somatic.vcf.gz -a GRCh38 -o ~/tmp/somatic.vep.vcf.gz
Possible precedence issue with control flow operator at /home/alanh/miniconda3/envs/vep/lib/site_perl/5.26.2/Bio/DB/IndexedBase.pm line 805.

-------------------- EXCEPTION --------------------
MSG: ERROR: No cache found for homo_sapiens, version 103

STACK Bio::EnsEMBL::VEP::CacheDir::dir /home/alanh/miniconda3/envs/vep/share/ensembl-vep-103.1-2/modules/Bio/EnsEMBL/VEP/CacheDir.pm:328
STACK Bio::EnsEMBL::VEP::CacheDir::init /home/alanh/miniconda3/envs/vep/share/ensembl-vep-103.1-2/modules/Bio/EnsEMBL/VEP/CacheDir.pm:227
STACK Bio::EnsEMBL::VEP::CacheDir::new /home/alanh/miniconda3/envs/vep/share/ensembl-vep-103.1-2/modules/Bio/EnsEMBL/VEP/CacheDir.pm:111
[...]

PR #970, opened 2021-04-16
--cache optionally takes parameter for directory

alanhoyle:
If the user specifies `--cache PATH`, **PATH** is used as the default value for the following other options:
   `--dir`, `--dir_cache`, `--dir_plugins`

The `PATH` is optional, and this should be backwards-compatible with existing command lines that include `--cache` as VEP does not appear to accept bare ARG values.  

This addresses issue #966.  I suspect many other people have tried `--cache $PATH` instead of `--cache --dir $PATH`

This could also possibly be implemented by putting logic into `AnnotationSourceAdaptor.pm` and `Runner.pm` instead of the top-level `vep`.  

This also deletes trailing whitespace characters on several lines.

### DIFF
--- a/vep
+++ b/vep
@@ -223,18 +223,13 @@ $config->{database} ||= 0;
 # If the optional parameter is supplied for --cache,
 #   AND that parameter is a directory
 #   AND --dir|dir_cache|dir_plugins is not also defined,
-#   set --dir|dir_cache|dir_plugins to the value given for --cache.
+#   set --dir|dir_cache to the value given for --cache.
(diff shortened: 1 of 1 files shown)


at7:
Thank you or this PR. We are going to review it in the coming weeks and let you know if we have any feedback. Eventually, this PR needs to be against the master branch in order for it to be part of the upcoming 105 release branch.

Best wishes,
Anja

alanhoyle:
I'm going to close this PR and redo the commit against master.

PR #975, opened 2021-04-21
--cache optionally takes parameter for directory

alanhoyle:
This is a replacement PR  for #970.  The changes are now implemented against master.  

If the user specifies `--cache PATH`, **PATH** becomes equivalent to `--dir PATH` which is used as the default value for `--dir_cache`, and in a modified form equivalent to `--dir_plugins PATH/Plugins` 

If the other options (`--dir`, `--dir_cache`, or `--dir_plugins`) are specified, they override whatever is set by the `--cache` option.  

**PATH** is optional, and this is backwards-compatible with existing command lines that include `--cache`. VEP does not appear to accept bare ARG values, so there should not be any unlucky clashes.

If no **PATH** is specified, the option behaves as before and simply used the defaults or other options as specified. 

This addresses issue #966.  I suspect many other people have tried `--cache $PATH` instead of `--cache --dir $PATH`

This could also possibly be implemented by putting logic into `AnnotationSourceAdaptor.pm` and `Runner.pm` instead of the top-level `vep`.  

The PR also removes extraneous whitespace characters on several lines.


at7:
Thank you

at7:
I can see the benefits of your code update. However, the VEP code base is complex and throughout the code the --cache option is treated as a flag. Overloading it with also adding a path option will lead to conflicting interpretations in other parts of the code which makes it easier for bugs to be introduced. We do have dedicated parameters for specifying cache and plugin directories and the default path for looking up the data is $HOME/.vep/".

alanhoyle:
I have added a few lines of code in a new commit that should address your concern:  

* If `--cache` is omitted, the behavior is exactly the same as before:  $config->{cache} is not defined
* if `--cache` alone is included:  $config->{cache} == 1
* if `--cache /path/to/dir`  is included, the value is used for as I want (--dir/--dir_cache/--dir_plugins as above), and then sets `$config->{cache} = 1` which matches old behavior
* if `--cache NOT_A_DIR` is not a directory, it throws a warning about it not being a directory, and then sets `$config->{cache} = 1` which matches old behavior

nuno-agostinho:
Hey @alanhoyle, thank you for your contribution and sorry for the delay in replying!

Upon discussing this with the team, we agree with using `--cache [dir]` based on your PR with some changes to the code:

1. Any changes done to CLI arguments should be consistent with using INI configuration files. As such, I moved your changes to the Config.pm file.
2. I had to modify the order of VEP's loading of config from arguments and INI files to make this easier to implement. This will make future changes easier, as well.
3. We think that adding a directory to `--cache` should only affect the argument `--dir_cache`, it is not intuitive to change the directory of plugins, etc. based on a directory to `--cache`.

Hope that you agree with our changes, but feel free to comment.

@diegomscoelho Could you please test this PR, including my changes? For instance, run VEP with:
- `--cache [dir]`
- `--cache --dir_cache [dir]`
- `--cache --dir_cache [dir]` (should prioritise `--dir_cache`)
- Using a INI config file with cache defined with a directory

We may still need to add new changes to the documentation to make it clear of this new use-case for new and existing VEP users.

Thanks!

Cheers,
Nuno

alanhoyle:
Thanks all! Appreciate this.

Issue #958, opened 2021-03-24
option --port doesn't seem to work


Sanger-ad7:
## Describe the issue
Our local mirror of ensembldb.ensembl.org has both assemblies, GRCh37 and GRCh38, available on port 3306 but when the port is specified via the "--port" option and the assembly is set to GRCh37 vep seems to ignore the port option and use the default port, 3337
It does, however, work properly when the database connection details are supplied via a config file
e.g 
       host  ensembldb-mirror
       port  3306
       assembly GRCh37
       db_version 102

### System
docker container  ensemblorg/ensembl-vep:release_102.0

### Full VEP command line
```
vep \
--fasta /opt/vep/.vep/homo_sapiens/102_GRCh37/Homo_sapiens.GRCh37.75.dna.primary_assembly.fa.gz \
--database \
--host ensembldb-mirror \
--port 3306 \
--assembly GRCh37 \
--db_version 102 \
--format vcf \
--dir_plugins /opt/vep/.vep/Plugins \
-i /opt/vcf/example.vcf.gz 
```

### Full error message
MSG: Cannot connect to the Ensembl MySQL server at ensembldb-mirror:3337; check your settings & DBI error message: Can't connect to MySQL server on 'ensembldb-mirror' (111)

at7:
Hi,
thank you for this report. It is true that the port number is hard coded when using the GRCh37 assembly. We will fix this for our next release 104 which is due in April. I will let you know when the release is out here.

Best wishes,
Anja

at7:
Hello,
Our latest release 104 is now out which contains a fix for this problem. Please update to the latest version and check that the problem is now resolved. Please also let me know if there are any further problems.

Best wishes,
Anja

PR #976, opened 2021-04-22
fix condition for setting grch37 port

at7:
Providing a different port number for GRCh37 databases was ignored because of small bug in the condition statement.
This was reported here https://github.com/Ensembl/ensembl-vep/issues/958
It can be tested with our GRCh37 mirror server.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Runner.pm
+++ b/modules/Bio/EnsEMBL/VEP/Runner.pm
@@ -114,7 +114,7 @@ sub init {
   # log start time
   $self->stats->start_time();
 
-  if(defined($self->param('assembly')) && lc($self->param('assembly')) eq 'grch37' && defined($self->param('database') && !defined($self->param('port')))) {
+  if(defined($self->param('assembly')) && lc($self->param('assembly')) eq 'grch37' && defined($self->param('database')) && !defined($self->param('port'))) {
     $self->param('port', 3337);
   }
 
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/VEP/Runner.pm
```diff
@@ -114,7 +114,7 @@ sub init {
   # log start time
   $self->stats->start_time();
 
-  if(defined($self->param('assembly')) && lc($self->param('assembly')) eq 'grch37' && defined($self->param('database') && !defined($self->param('port')))) {
+  if(defined($self->param('assembly')) && lc($self->param('assembly')) eq 'grch37' && defined($self->param('database')) && !defined($self->param('port'))) {
```
dglemos, line 117: It fixes the issue. However, if we specify the assembly grch37 without the port it doesn't go into the condition. 
When it gets to this point, the port is defined `3306` and the check `!defined($self->param('port'))` fails. 
Example: 
perl vep --id "rs699" --database --db_version 103 --assembly GRCh37

-------------------- EXCEPTION --------------------
MSG: ERROR: Assembly version specified by --assembly (GRCh37) and assembly version in coord_system table (GRCh38) do not match

at7: Thank you for spotting this!

dglemos (approved):

aparton:
merged into release/104 and master

Issue #974, opened 2021-04-21
Half "empty" html stats file


jd3234:
## Describe the issue

The html file is generated and the top part (description and overview table) are visible (html file attached at the end). But the pie and bar charts are not visible. 

![grafik](https://user-images.githubusercontent.com/32764481/115542798-ef687700-a2a0-11eb-8f4d-88eda068a936.png)

Upon inspection of the html code I found that a variable is set that contains the name of chromosomes. In my case there is a `-` in the contig names which seems to cause the problem.

Original variable in html (not working): `var chr_HA_NAME_hq-v1_chr01_area = drawArea`  
Variable without `-` (working): `var chr_HA_NAME_hq_v1_chr01_area = drawArea`

Example of chromosome names in fasta file/annotation:
```
>HA_NAME_hq-v1.chr01
ACTGAC
>HA_NAME_hq-v1.chr02
ACGTAG
```

### System
- VEP version: 10.3 (conda)
- VEP Cache version: custom
- Perl version: 
- OS: Ubuntu
- tabix installed ? yes

### Full error message
There are no error/warning messages.

### Data files (if applicable)
I am attaching the non working html file as txt file: 
[STDOUT_summary.html.txt](https://github.com/Ensembl/ensembl-vep/files/6350246/STDOUT_summary.html.txt)

at7:
Hello,
thank you for reporting this problem and thank you also for pointing straight to the cause. We will be working on a bug fix and should be able to include the fix in our upcoming release 104. I will keep you posted here about our progress.

Best wishes,
Anja

at7:
Hello,
Our latest release 104 is now out which contains a fix for this problem. Please update to the latest version and check that the problem is now resolved. Please also let me know if there are any further problems.

Best wishes,
Anja

jd3234:
Sorry forgot to answer. Tested and fully working! Thanks! :)

PR #977, opened 2021-04-22
replace dash in var name

at7:
A hyphen in a chromosome name which is used for creating the javascript variable name was breaking the html display. Hyphens are not allowed in javacscript names. My script for reproducing and testing can be found here PANDA/anja/vep_issue_974.
Issue was reported here #974


dglemos (approved):

dglemos:

PR #986, opened 2021-05-11
OutputFactory json - do not convert gene symbol to integer

dglemos:
The SpliceAI json output:
```
"spliceai": {
 "DP_DL":7,
 "DS_AL":0,
 "DP_AG":19,
 "DS_DL":0,
 "SYMBOL":"NANS",
 "DS_AG":0.01,
 "DP_AL":1,
 "DP_DG":-4,
 "DS_DG":0
}
```
The OutputFactory/JSON tries to numberify the SYMBOL string. For gene "NANS" on REST, the string is converted to `-nan`
Ticket: ENSVAR-4087

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory/JSON.pm
@@ -116,6 +116,7 @@ my %NUMBERIFY_EXEMPT = (
   'gene_id' => 1,
   'gene_symbol' => 1,
   'transcript_id' => 1,
+  'SYMBOL' => 1
 );
 
 my @LIST_FIELDS = qw(


at7 (commented):
The SYMBOL key comes from the SpliceAI plugin. The JSON output factory tries to numberify SYMBOL because it was not in the exempt hash. It then updates the gene NANS into a number -nan, which in REST causes the error "SyntaxError: JSON.parse: unexpected non-digit".
Adding SYMBOL to the exempt hash works.
But in order to avoid adding keys from plugins into the exempt hash I wonder if we could update numberify and ignore testing for looks_like_number if nan is in the value (https://www.perl.com/article/validating-untrusted-input-numbers/). But that would require more code updates and should be a jira ticket for future developments.

So I'm happy with the PR. But:
-- could you please do the PR separately against
  --master
  -- release/104 with sub version update?
-- add a comment that SYMBOL comes from the SpliceAI plugin
-- create a jira ticket to review the numberify method?

Thank you!

dglemos:
@at7 thank you for the comments. I created two PR against master and release/104 as requested. I also created a ticket for future development (review numberify).

dglemos:
Replaced by #990 and #991

PR #993, opened 2021-05-25
Select correct codon table for MT chromosome

dglemos:
VEP does not use mitochondrial codon table when source is GTF or GFF. 
ENSVAR-4061


at7 (approved):

at7:
Could you please update VEP_SUB_VERSION?

modules/Bio/EnsEMBL/VEP/Constants.pm
```diff
@@ -53,7 +53,7 @@ use warnings;
 use base qw(Exporter);
 
 our $VEP_VERSION     = 104;
-our $VEP_SUB_VERSION = 0;
+our $VEP_SUB_VERSION = 2;
```
at7, line 56: This should be 3 now and we need it to create a new tag.

at7 (approved):

at7:
merged into master and release/104

PR #1014, opened 2021-06-24
Variant Recoder to report MANE transcripts

dglemos:
Report MANE select transcripts: for each transcript it report HGVS genomic, HGVS coding and HGVS protein. 
ENSVAR-4074


aparton (changes requested):
If I use the `--fields` option, I get empty values:

`perl variant_recoder -id rs699 --db_version 104 --pretty --mane_select --fields spdi,mane_select`

gives empty values when the mane select is correctly populated with the fields option removed

dglemos:
`--mane_select` retrieves the values from the hgvs notations. When mane_select is on, `hgvsg`, `hgvsc` and `hgvsp` also need to be on. It works now.

aparton:
Merged into release/105 and master

PR #1021, opened 2021-07-09
Remove column 'Existing_variation' from VEP default output

dglemos:
The column 'Existing_variation' is returned as default in VEP output however, it only returns data if --check_existing (or other option that switches on --check_existing) is used. 
As no data is returned by default, the column should be removed. 

Details in ENSVAR-4250


dglemos:
Deleting the column from the default columns is going to change the output column order. It's safer if we let the columns as they are and update the documention to make it clear that 'Existing_variation' is only populated if run vep with option to populated it.

PR #1026, opened 2021-07-20
GA4GH VRS Allele object to variant recoder

helensch:
Adds option to output GA4GH VRS Allele object in Variant Recoder and VEP
For further info see: 
ENSVAR-4092 Variant Recoder
ENSVAR-4186 VEP

Requires https://github.com/Ensembl/ensembl-variation/pull/739

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -396,8 +396,7 @@ our %INCOMPATIBLE = (
   tab         => [qw(vcf json)],
   individual  => [qw(minimal)],
   check_ref   => [qw(lookup_ref)],
-  check_svs   => [qw(offline)],
-  ga4gh_vrs   => [qw(vcf tab txt)]
+  check_svs   => [qw(offline)]
 );
 
 # deprecated/replaced flags
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -381,6 +381,7 @@ our %REQUIRES = (
   phyloP    => [qw(ucsc_assembly)],
   phastCons => [qw(ucsc_assembly)],
   custom_multi_allelic => [qw(custom)],
+  ga4gh_vrs => [qw(json)]
 );
 
 # incompatible options
@@ -396,6 +397,7 @@ our %INCOMPATIBLE = (
   individual  => [qw(minimal)],
   check_ref   => [qw(lookup_ref)],
   check_svs   => [qw(offline)],
+  ga4gh_vrs   => [qw(vcf tab txt)]
 );
 
 # deprecated/replaced flags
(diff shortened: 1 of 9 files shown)


modules/Bio/EnsEMBL/VEP/Config.pm
```diff
@@ -396,6 +397,7 @@ our %INCOMPATIBLE = (
   individual  => [qw(minimal)],
   check_ref   => [qw(lookup_ref)],
   check_svs   => [qw(offline)],
+  ga4gh_vrs   => [qw(vcf tab txt)]
```
aparton, line 400: With the change on line 384, adding this list to %INCOMPATIBLE isn't needed

modules/Bio/EnsEMBL/VEP/Constants.pm
```diff
@@ -172,6 +173,7 @@ our %FIELD_DESCRIPTIONS = (
   'HGVSp'              => 'HGVS protein sequence name',
   'HGVSg'              => 'HGVS genomic sequence name',
   'SPDI'               => 'Genomic SPDI notation',
+  'GA4GH_VRS'          => 'GA4GH VRS object',
```
aparton, line 176: As this option is only available with JSON output, then the Constants.pm updates are unnecessary as we wont be printing header lines. However, if we might extend this functionality to other VEP output formats in the future, then these would be needed. 

If we do decide to keep them for future-proofing purposes, it may be worth adding more detail to the field description

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -1286,6 +1287,17 @@ sub VariationFeatureOverlapAllele_to_output_hash {
     }
   }
 
+  # ga4gh_vrs
+  if ($self->{ga4gh_vrs}) {
+    $vf->{_ga4gh_spdi_genomic} = $vf->spdi_genomic();
+
+    if (my $ga4gh_spdi = $vf->{_ga4gh_spdi_genomic}->{$hash->{Allele}}) {
+      if ($ga4gh_spdi =~ /^NC/) {
```
aparton, line 1295: I'm not fully clued up on the VRS spec so this might be a silly question, but are we comfortable with how this works on patches?

aparton (commented):
Looks good. Runs successfully with all of my test variants, tests pass when the associated ensembl-variation PR is included.

Minor comments on Config/Constants attached

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
helensch: For the initial implementation, VEP and Variant Recoder will only  output GA4GH VRS object for NC locations. For patches no GA4GH VRS is output even though SPDI is available. See output from VEP runs for rs1210229297

modules/Bio/EnsEMBL/VEP/Config.pm
helensch: Thanks, the ga4gh_vrs has been removed from the incompatible list

modules/Bio/EnsEMBL/VEP/Constants.pm
helensch: Thanks, I have removed the GA4GH_VRS info from the Constants.pm

helensch:
@aparton thanks for the review and flagging the options and info not needed.
Please could you review the PR again.

PR #1068, opened 2021-10-26
Include missing values in JSON output

dglemos:
'used_ref' and 'given_ref' are missing from JSON output when the value is '-'
More details: ENSVAR-4492


sarahhunt:
merged to release/106 and master

PR #1080, opened 2021-11-18
Fix VEP frequency filtering - match alleles

dglemos:
Details: ENSVAR-4529


jamie-m-a:
Code looks good, test ran fine, impact on run time seems minimal.

PR #1115, opened 2022-02-18
feat: Removal of ESP + Include Gnomad 3.2.1 in DumpVEP

diegomscoelho:
Jira CARD: [Here](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4051)

## Description

The NHLBI-GO Exome Sequencing Project (ESP) samples have been included in the gnomAD exomes set so can be removed from the VEP cache. Besides that we must need to include Gnomad 3.2.1 genomes INFO + GMAF from 1000 genomes to this dump. This alterations must be aligned with vep alterations to get new cache data input. eHive pipeline should be affected by these alterations.

## Technical description

1) It was necessary to include this regex replace to avoid `chr` be included by default in some cases. [code](https://github.com/Ensembl/ensembl-vep/pull/1115/files#diff-2c19daba2931ed372d883093c3e1fb1331b98b00a59647065bb31cc28e69f231R286-R287)
2) `af_gnomad` was divided in two commands: `af_gnomade` and `af_gnomadg` for gnomAD exomes and genomes, respectively.
3) `AF` field is no more taken from variation `minor_allele_freq` col. This information comes directly from stored `AF` from 1000 genomes dump file.
4) Dump_vep pipeline stores gnomAD exomes, gnomAD genomes and also GMAF from 1000 genomes, and ESP information were removed from default.
4) Tests were updated as some input test fields. It would be good to include tests to new fields.

old version -> gnomAD_AFR_AF
newer version -> gnomADe_AFR_AF

## Test

1) Make sure new updates on pipeline and new data names make sense.
2) Try using `-af_gnomade` flag with and without this branch PR.
3) Try using `-af_gnomadg` flag.
4) eHive results alterations can be found in JIRA Card.

## Necessary for
https://github.com/Ensembl/public-plugins/pull/545
https://github.com/diegomscoelho/ensembl-vep/pull/1


diegomscoelho:
Hi @jamie-m-a,

Removed submodule from file changes.

sarahhunt:
It would be better to remove ESP and add gnomAD genomes in the same release to avoid frequent changes. Anyone using ESP in their pipelines then has the option to switch to gnomAD as they make the update to remove ESP.

diegomscoelho:
Added `--af_gnomad` again as a flag to get gnomAD exomes data.
Important: INFO fields were updated:
old version -> gnomAD_AFR_AF
newer version -> gnomADe_AFR_AF

diegomscoelho:
@jamie-m-a I had added frequency files checking in last two commits. Basically it will only check listed chromosomes, but if does not exist it will `kill` DumpVEP.

jamie-m-a (approved):

PR #1174, opened 2022-05-03
Renaming master to main for ensembl-variation

likhitha-surapaneni:
Renaming master to main for files dependent on ensembl-variation repo


nuno-agostinho (approved):
Seems fine, thank you!

PR #1182, opened 2022-05-17
Annotation of MT transcripts in RefSeq

likhitha-surapaneni:
[ENSVAR-4484](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4484)
This PR attempts to fix the pattern matching for MT transcripts

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationType/Transcript.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationType/Transcript.pm
@@ -210,6 +210,7 @@ sub filter_transcript {
     (
       # we only want RefSeq transcripts e.g. NM_12930,
       # or 4540 for MT transcripts
+      # or rna-TRNP for MT transcripts
       (
         $self->{source_type} eq 'refseq' &&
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/VEP/AnnotationType/Transcript.pm
```diff
@@ -210,9 +210,10 @@ sub filter_transcript {
     (
       # we only want RefSeq transcripts e.g. NM_12930,
       # or 4540 for MT transcripts
+      # or rna-TRNP for MT transcripts
       (
         $self->{source_type} eq 'refseq' &&
-        !(($t->stable_id || '') =~ /^[A-Z]{2}\_\d+|^\d{4}$/ || 
+        !(($t->stable_id || '') =~ /^[A-Z]{2}\_\d+|^\d{4}$|^rna-[A-Z]{3,4}\d+/ ||
```
sarahhunt, line 216: Thanks @likhitha-surapaneni . There are transcripts without numbers in name - like rna-TRNV- so both pattern matches need to end with \d{0,1} (to permit no digits) rather than \d+

nuno-agostinho:
Please check whether this should go for the new `postreleasefix/108` instead.

jamie-m-a (approved):
Regex now matches both circumstances

Issue #1178, opened 2022-05-12
combination of --check_ref --dont_skip fails to catch problematic lines


ghost:
## Describe the issue
With the file below, --check_ref and --dont_skip fail to produce expected string "failed" in the "CHECK_REF" column. Variant where REF is AAT does not match the reference, or does it?

### System
- VEP version: 105
- VEP Cache version: 105
- Perl version: 5.32.1
- OS: Debian
- tabix installed: yes

### Full VEP command line
```
./vep  -input_file /opt/vep/.vep/input/test.vcf --offline --cache --cache_version 105 --dir_cache /opt/vep/.vep/reference --refseq --hgvs --hgvsg --fasta /opt/vep/.vep/reference/Homo_sapiens.GRCh37.dna.primary_assembly.fa.gz --format vcf --vcf --check_ref --dont_skip --force_overwrite --no_stats -o stdout > /opt/vep/.vep/output/test_VEP_RefSeq.vcf
```

### Input
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	S21
chr13	32900399	.	AAT	A	.	PASS		GT:DP:AF	0/1::0.0138313
chr13	32900399	.	TAT	A	.	PASS		GT:DP:AF	0/1::0.0138313
chr13	32900399	.	GAT	A	.	PASS		GT:DP:AF	0/1::0.0138313
chr13	32900399	.	CAT	A	.	PASS		GT:DP:AF	0/1::0.0138313

### Output
##fileformat=VCFv4.1
##VEP="v105" time="2022-05-12 14:56:46" cache="/opt/vep/.vep/reference/homo_sapiens_refseq/105_GRCh37" ensembl-variation=105.ac8178e ensembl-io=105.2a0a40c ensembl=105.525fbcb ensembl-funcgen=105.660df8f 1000genomes="phase3" COSMIC="92" ClinVar="202012" HGMD-PUBLIC="20204" assembly="GRCh37.p13" dbSNP="154" gencode="GENCODE 19" genebuild="2011-04" gnomAD="r2.1" polyphen="2.2.2" refseq="2020-10-26 17:03:42 - GCF_000001405.25_GRCh37.p13_genomic.gff" regbuild="1.0" sift="sift5.2.2"
[...]

PR #1183, opened 2022-05-18
Check ref bug

olaaustine:
[ENSVAR-4844](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4844)
#1178


jamie-m-a:
This looks fine, but as Sarah mentioned, might be worth checking impact on speed. And as Diego mentioned, maybe adding a test that would have caught this issue originally.

jamie-m-a (approved):
Thanks for updating with testing details, looks good!

PR #1190, opened 2022-05-23
Creating temp_dir before running pipeline should be mandatory

nakib103:
Die if temp_dir creation fails. In some cases, temp_dir creation may fail, for example, in case it needs to be created recursively. We currently made it mandatory pre-requisite to create temp_dir before running pipeline instead of handling it on the code -

https://www.ebi.ac.uk/seqdb/confluence/display/EV/VEP+dumps#VEPdumps-Pre-requisites

JIRA - https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4860


diegomscoelho (approved):
LOTM!

PR #1198, opened 2022-05-31
Patch schema and test databases for 108

nuno-agostinho:
[ENSVAR-4813](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4813)


jamie-m-a (approved):
Looks good to me - thanks for fixing the consequences in transcript_variation

PR #1205, opened 2022-06-16
fix: broken tests due to altering --spdi usage

diegomscoelho:
## Description

An alteration in https://github.com/Ensembl/ensembl-variation/pull/822 crashed a few tests in `ensembl-vep` repo. This PR fix those old tests.

## Tests

1) Travis need to work fully.


diegomscoelho:
Ending this test fix and solving in https://github.com/Ensembl/ensembl-variation/pull/823

PR #1214, opened 2022-06-28
catch unexpected extra command-line parameters at run time

willmclaren:
GetOptions doesn't seem to care about anything left hanging in `@ARGV`, so if a user inadvertently adds an additional parameter (e.g. adding a value to a param that doesn't ask for one like `--cache`) this will get left in `@ARGV`.

In most cases this will be harmless, but if using the LOFTEE plugin the [score3.pl](https://github.com/konradjk/loftee/blob/master/maxEntScan/score3.pl) script tries to open `$ARGV[0]` and dies with a difficult to trace error message.

I don't see any test file for `vep` itself; the following command will cause the die to trigger:

```vep --cache 99```

### DIFF
--- a/vep
+++ b/vep
@@ -217,6 +217,8 @@ GetOptions(
   'debug',                   # print out debug info
 ) or die "ERROR: Failed to parse command-line flags\n";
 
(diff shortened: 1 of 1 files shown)


dglemos (approved):
@willmclaren thanks for the update!

PR #1216, opened 2022-07-12
Parsing pos/len in json with --total_length argument

likhitha-surapaneni:
[ENSVAR-4993](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4993)

This PR attempts to replace `looks_like_number` with regex to account for the pattern pos/len


olaaustine (approved):
Looks good!

PR #1230, opened 2022-08-01
Annotation from file GFF to skip miRNA

dglemos:
Details in [ENSVAR-3149](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-3149)


diegomscoelho:
Adjustments are working for this case. Running:

```sh
vep \
--i  G92F_CKDN200007804-1A_trimmed_merged_sorted.vcf \
--cache --no_stats --vcf \
--output_file result.vcf \
--gff GCF_009914755.1_T2T-CHM13v2.0_genomic.gff.gz \
--offline --cache_version 107 \
--dir_cache <CACHE_DIR> \
--assembly GRCh38 --force_overwrite \
--fasta Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz \
```

I have two questions:

1) Are we going to deal only with `miRNA` as type?
2) Should we not include a die in case of type is a non-predicted one? - This will avoid new types breaking without being noticed.

diegomscoelho (approved):
LGTM! Tested again and it's working fine. Triggered test again (Perl 5.10 had not allowed access to ensembl-variation - probably some bug on test). If Travis say it's OK I will merge.

PR #1251, opened 2022-09-01
Variant recoder MANE select only available for human

dglemos:
[ENSVAR-4364](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4364)


nakib103 (approved):
Looks Good! ~

PR #1253, opened 2022-09-06
fix: use of get_all_overlapping_Peaks method

diegomscoelho:
JIRA: [Card](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5135)

## Description

While running last VEP Dump this bug was found:

```sh
Can't use an undefined value as an ARRAY reference at /hps/software/users/ensembl/repositories/enseven/ensembl-vep/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/RegFeat.pm line 224.
```

## Test

1) Just need to make sure that Vep Dump it's working with new function for RegFeat.pm


Gurpreet-Ghattaoraya:
Tested this pull request: I was able to recreate the job failure with the original code.
The same job ran sucessfully with the change in this pull request.

Gurpreet-Ghattaoraya (approved):

dglemos:
Merged into release/108 and main

PR #1262, opened 2022-09-16
Fix gzip error for compressed input file

nakib103:
[ENSVAR-4595](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4595) 

If `gzip` fails we can try the next with `IO::Uncompress::Gunzip` instead of dying. It will add greater reliability without compromising performance.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Utils.pm
+++ b/modules/Bio/EnsEMBL/VEP/Utils.pm
@@ -430,17 +430,14 @@ sub get_compressed_filehandle {
   my $fh;
   if($CAN_USE_PERLIO_GZIP && !$multi) {
     eval q{ open $fh, "<:gzip", $file; };
-    warning("Opening $file failed using PerlIO::gzip\nERROR: $@\nTrying other methods..\n") if $@;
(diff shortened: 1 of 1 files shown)


diegomscoelho (commented):
I think all those warnings can be quite confusing and misleading. If any of gzip method works, I would not print anything at all. For example, I'm running `vep` and always having this message (although output is working well):

```sh
-------------------- WARNING ----------------------
MSG: Opening small.vcf.gz failed using system gzip
ERROR: Illegal seek
Trying other methods..

FILE: EnsEMBL/VEP/Utils.pm LINE: 438
CALLED BY: EnsEMBL/VEP/Parser.pm  LINE: 236
Date (localtime)    = Mon Sep 26 14:03:53 2022
Ensembl API version = 109
---------------------------------------------------
```

When you finish updating tests, please ask my review again.

nakib103:
It is not exactly solving the problem. Because the warning `gzip: stdout: Broken pipe` does not occur at the open command rather at some point while reading the file. So, trying to fix when the file_handles are created would be wrong approach.

Issue #981, opened 2021-04-29
Suggestion: Removal of cpan cache in Dockerfile


sigven:
Hi,

I noticed that the VEP Docker image contains a cache that, when deleted, will reduce the overall size of the image with some ~125-130Mb.  I simply added `rm -rf /root/.cpanm` at the end of [this RUN command](https://github.com/Ensembl/ensembl-vep/blob/release/103/docker/Dockerfile#L140).  My understanding is that this cache is not used during any analysis, so might want to include the statement that deletes it.

kind regards,
Sigve

helensch:
Hi @sigven

Thank you for your suggestion to remove the cpan cache. We will review for inclusion in the next update.

Regards
Helen

nuno-agostinho:
Hi @sigven, thanks for your suggestion!

We removed the CPAN cache from the Dockerfile and this should be reflected in our future Docker images.

Feel free to suggest other improvements you may think relevant.  Hope you have a nice day!

Cheers,
Nuno

PR #1273, opened 2022-10-10
Improve VEP Docker image

nuno-agostinho:
* [ENSVAR-4082](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4082) and #981
  * Remove `.cpanm` cache for root (not required to run VEP)
  * Reduces image size from 787MB to 665MB
* [ENSVAR-5183](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5183)
  * Automatically skip update checks for VEP and Ensembl API when running `INSTALL.pl`
* [ENSVAR-5184](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5184)
  * Change default working directory to `/data`

### Testing

1. Build a Docker image with the Dockerfile from this PR, e.g., `docker build -t vep .` (replace `ENV BRANCH main` with `ENV BRANCH release/109` in the Dockerfile, otherwise you won't be able to test skipping of update checks when running `INSTALL.pl`)
2. Run `docker run -ti vep` and check if `pwd` is `/data`
3. When running simply the command `INSTALL.pl`, script should automatically skip the first question about updating VEP and Ensembl API
4. Pull latest VEP Docker image from Docker Hub
5. Check if image size is reduced using `docker images`
6. You can try running VEP in both images and comparing their output, it should be the same for the same command

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -150,6 +150,7 @@ RUN ensembl-vep/travisci/build_c.sh && \
     /usr/sbin/update-locale LANG=$LANG_VAR && \
     # Copy htslib executables. It also requires the packages 'zlib1g-dev', 'libbz2-dev' and 'liblzma-dev'
     cp $HTSLIB_DIR/bgzip $HTSLIB_DIR/tabix $HTSLIB_DIR/htsfile /usr/local/bin/ && \
+    # Delete cache from cpanm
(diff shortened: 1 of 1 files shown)


likhitha-surapaneni (approved):
Looks good to me

PR #1274, opened 2022-10-12
Hotfix: accept cache argument without value

nuno-agostinho:
This hotfix allows running VEP using `--cache` without a value, e.g.: `vep -i variants.vcf --cache`.


diegomscoelho (approved):
LGTM! Tested with and without alteration.

PR #1280, opened 2022-10-19
Fix `--mirna` returning miRNA structures for all ncRNAs

### LINKED ISSUE #1278: VEP --mirna annotates more than miRNAs
## Describe the issue
Hello VEP devs,
I am trying out the --mirna option to VEP, which according to the doc "Reports where the variant lies in the miRNA secondary structure". This sounds interesting, but I would expect it to only annotate variants hitting microRNAs (BIOTYPE==miRNA). However in my tests I get annotations in the "miRNA" field for a variety of BIOTYPEs including snoRNA and snRNA.
For example, running VEP --miRNA on rs150964208 , I obtain an annotation for the snoRNA SNORA15 that contains "miRNA_loop" in the miRNA field. This seems wrong?

### System
- VEP version: 107
- VEP Cache version: homo_sapiens_vep_107_GRCh38
- Perl version: perl-5.16.3-299.el7_9.x86_64
- OS: CentOS 7
- tabix installed ? yes

### Full VEP command line
The result is obtained using the VEP web interface, checking the "miRNA structure" box.

### Full error message
None, but the result is unexpected.

nuno-agostinho:
Fixes #1278:  when running `--mirna`, it is currently displaying miRNA structures for any ncRNA, even if they are not miRNA.

Test these variants using `vep --id [rsID] --mirna --cache [cache_dir]`:
* rs569478349 that falls within a miRNA (`ENST00000577608`) and should return `miRNA=miRNA_stem`
* rs150964208 that falls within a ncRNA (that is not miRNA, `ENST00000384439`) and shouldn't return any miRNA structure (it was returning `miRNA=miRNA_loop` before)

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
@@ -1487,7 +1487,9 @@ sub BaseTranscriptVariationAllele_to_output_hash {
   }
 
   # miRNA structure
-  if($self->{mirna} && (my ($mirna_attrib) = grep {$_->code eq 'ncRNA'} @attribs)) {
+  if($self->{mirna} && $tr->biotype eq 'miRNA' &&
+     (my ($mirna_attrib) = grep {$_->code eq 'ncRNA'} @attribs)) {
+
     my ($start, $end, $struct) = split /\s+|\:/, $mirna_attrib->value;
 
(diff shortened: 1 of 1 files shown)


olaaustine (approved):
Looks good, works as expected. Thanks @nuno-agostinho.

PR #1282, opened 2022-10-20
Some re-factoring on nextflow vep

nakib103:
- New option `--output_prefix` to overwrite output file name
- Do not try to create index file if already present
- It has been always taking chromosomes from the `params.chros` - by default comment out the `param.chros` in config (reverted).
- The tabix index file is mandatorily generated, it is better to use `tabix` instead of `bcftools` to read chromosomes from the file because that will remove dependency from the header. So even if there is no proper header in the input vcf we will get the chromosomes.

### DIFF
--- a/nextflow/nf_config/nextflow.config
+++ b/nextflow/nf_config/nextflow.config
@@ -37,3 +37,4 @@ profiles {
 params.chros_default = "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y,MT"
 //params.chros_file = "$PWD/examples/clinvar-testset/chros.txt"
 params.vep_config = "$PWD/nf_config/vep.ini"
+params.output_prefix = ""
--- a/nextflow/nf_config/nextflow.config
+++ b/nextflow/nf_config/nextflow.config
@@ -34,6 +34,6 @@ profiles {
   }  
 }
 
-params.chros = "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y,MT"
+params.chros_default = "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y,MT"
 //params.chros_file = "$PWD/examples/clinvar-testset/chros.txt"
 params.vep_config = "$PWD/nf_config/vep.ini"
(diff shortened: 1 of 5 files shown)


nextflow/nf_config/nextflow.config
```diff
@@ -34,6 +34,7 @@ profiles {
   }  
 }
 
-params.chros = "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y,MT"
+params.chros_default = "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,X,Y,MT"
```
nuno-agostinho, line 37: There's no need to prefix this with `param.` because we don't want users to use `--chros_default` as an argument.

nakib103: Thanks @nuno-agostinho , updated 👍

nextflow/workflows/run_vep.nf
```diff
@@ -59,14 +59,18 @@ if(check_bgzipped.exitValue()){
...
-check_parsing.consumeProcessOutput(sout, serr)
-check_parsing.waitFor()
-if( serr ){
-  exit 1, "The specified VCF file has issues in parsing: $serr"
-}
+
 vcf_index = "${params.vcf}.tbi"
 
+if( !file(vcf_index).exists() ) {
+  check_parsing = "$params.singularity_dir/vep.sif tabix -p vcf -f $params.vcf".execute()
```
likhitha-surapaneni, line 66: The code currently assumes that the index file is always present. This part of code was originally placed to check if the index file was created without corruption, not to create an index file if not present. Is there a better way to implement this?

likhitha-surapaneni:
Thank you for the PR @nakib103 . Tested --output_prefix, default option for chromosome ids and parsing of chromosome id parsing with tabix. Works as expected. Can we also update the help text to show the updated parameters along with the README ?

nakib103:
> Thank you for the PR @nakib103 . Tested --output_prefix, default option for chromosome ids and parsing of chromosome id parsing with tabix. Works as expected. Can we also update the help text to show the updated parameters along with the README ?

@likhitha-surapaneni , I have updated the README.md with the new --output_prefix parameter.

nextflow/workflows/run_vep.nf
nakib103: Thanks for explaining - I have reverted this change. I think current approach works well. My problem in particular was related to write permission issue in the desired folder.

likhitha-surapaneni (approved):
Looks good to me. Thanks @nakib103

PR #1286, opened 2022-10-21
Hotfix INSTALL.pl URL downloads

### LINKED ISSUE #1284: INSTALL.pl can no longer install caches
Hi,
a recent commit replaced ftp with https everywhere:
https://github.com/Ensembl/ensembl-vep/commit/6185f08
This includes changing $CACHE_URL , $CACHE_URL_INDEXED and $FASTA_URL from ftp:// to https:// in INSTALL.pl, and this breaks INSTALL.pl: caches and FASTA files can no longer be installed.
Indeed, INSTALL.pl expects the URLs to begin with ftp, eg:
https://github.com/Ensembl/ensembl-vep/blob/0cc6b7c3471fc8e9ce9fd574156410e96475f141/INSTALL.pl#L1109

Simply reverting the INSTALL.pl part of commit 6185f08 solves the issue.

nuno-agostinho:
Fixes #1284: `INSTALL.pl` cache/FASTA downloads are breaking in `release/108` because they expect FTP URLs but we have changed them for HTTPS links. This PR fixes those issues.

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -401,6 +401,12 @@ sub update() {
 
 # CHECKS DIR SETUP AND PATHS ETC
 ################################
+
+sub is_url {
+  my $url = shift;
+  return $url =~ /^http/i;
+}
+
 sub check_default_dir {
   my $this_os =  $^O;
   my $default_dir_used;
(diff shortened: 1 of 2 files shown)


ntm:
Hi @nuno-agostinho ,
thanks for tackling this!
I don't think your fix is correct, however: you are forcing URLs to be http or https, but then you are still using Net::FTP to access the content. This only works today because an FTP server is actually listening on tcp/22 on the indicated host (ftp.ensembl.org), and although the URLs are specified with https:// , the code doesn't connect to tcp/443 (https). In other words the code is saying "naah I know these https:// URLs are wrong and should be ftp:// , let's connect to the servers with FTP". If you really want the URLs to be https, I think you need to use the https protocol to connect to them. That means probably writing some code to scrape the html pages eg https://ftp.ensembl.org/pub/release-108/variation/indexed_vep_cache/ , instead of the current $ftp->ls strategy.

nuno-agostinho:
Hi @ntm,

Thanks for your suggestion. This PR is intended as a quick fix for the current release and we currently have no plans of changing the way the FTP server is working. However, we do intend to improve this script for future releases of VEP.

Cheers,
Nuno

nakib103 (approved):
Tested and it works! LOTM ~

dglemos (approved):

nuno-agostinho:
Also merged to `main`, `release/109` and `postreleasefix/109`.

PR #1290, opened 2022-10-27
fix: Include full command line to all setups

diegomscoelho:
Jira: [Card](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5236)

## Description

Full command-line was not added to header if command-line wasn't used. Besides that, web VEP output included a white-space as a full command-line in header. This PR fixes both problems. Default values wasn't included on output.

## Test

1) Test `--config` setup;
2) Test VEP flags;
3) Web VEP (Sandbox in card).

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -476,15 +476,6 @@ sub new {
   # read environment variables with a VEP_ prefix, e.g. VEP_DIR_PLUGINS
   $self->read_config_from_environment($config);
 
-  my $config_command;
-
-  foreach (keys $config) {
-    next if $config->{$_} eq 0 || (ref($config->{$_}) eq "ARRAY" && @{$config->{$_}} == 0);
-    $config_command .= "--$_ $config->{$_} ";
-  }
-
-  $config->{full_command} = "vep $config_command" if !defined($config_command);
-
   # assign default port for GRCh37
   if (defined($config->{'assembly'}) && lc($config->{'assembly'}) eq 'grch37' && defined($config->{'database'}) && !defined($config->{'port'})) {
     $config->{'port'} = 3337;
@@ -495,6 +486,16 @@ sub new {
     $config->{dir_cache} ||= $config->{cache} if -d "$config->{cache}";
     $config->{cache} = 1;
   }
+
+  my $config_command;
+
+  for (keys($config)) {
+    next if $config->{$_} eq 0 || (ref($config->{$_}) eq "ARRAY" && @{$config->{$_}} == 0) || $_ eq "full_command";
+    
+    $config_command .= $config->{$_} eq 1? "--$_ "  : "--$_ $config->{$_} ";
+  }
+
+  $config->{full_command} = "vep $config_command";
 
   # set all other defaults
   foreach my $key(keys %DEFAULTS) {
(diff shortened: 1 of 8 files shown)


modules/Bio/EnsEMBL/VEP/Config.pm
```diff
@@ -487,6 +487,22 @@ sub new {
     $config->{cache} = 1;
   }
 
+  my $config_command;
+
+  my @skip_opts = qw(web_output host port stats_file user warning_file input_data);
+
+  foreach my $flag (sort keys %$config) {
+    my $value = $config->{$flag};
+    next if !defined($value) || $value eq 0 || (ref($value) eq "ARRAY" && @{$value} == 0) || grep { /$flag/ } @skip_opts;
```
nuno-agostinho, line 496: Is there a parameter where the user may be expected to pass a value of `0` that would be ignored by this `if` statement?

I don't think so, but just want to confirm.

modules/Bio/EnsEMBL/VEP/Config.pm
```diff
@@ -487,6 +487,22 @@ sub new {
...
+
+    $value = join(" --$flag ", @{$value}) if ref($value) eq "ARRAY";
```
nuno-agostinho, line 498: Do all flags share the same name between the internal flag and VEP argument?

modules/Bio/EnsEMBL/VEP/Config.pm
```diff
@@ -487,6 +487,22 @@ sub new {
...
+    $value =~ s/(\/[\w-]+?)+\//\[PATH\]\//g;
+    $config_command .= $value eq 1? "--$flag "  : "--$flag $value ";
```
nuno-agostinho, line 500: This makes sense if all the flags with a value of `1` are boolean. Are there any flags that receive 1 as a numeric value?

nuno-agostinho (commented):
Hey @diegomscoelho, thanks for your PR!

The PR seems to work fine in your web sandbox, as well as to print the expected arguments when using CLI flags, a INI configuration file and environment variables (e.g., `export VEP_PICK=1`).

Still, I put some comments that I would like to get your insight on. Thanks!

t/AnnotationSource_File_GFF.t
```diff
@@ -476,7 +476,7 @@ SKIP: {
 
   $runner->run;
   open IN, $test_cfg->{user_file}.'.out';
-  my @tmp_lines = <IN>;
+  my @tmp_lines = grep {!/^\#/} <IN>;
```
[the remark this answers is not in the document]
diegomscoelho: This alteration is no longer necessary, removing it.

modules/Bio/EnsEMBL/VEP/Config.pm
diegomscoelho: Four have possible numeric inputs:
```
--buffer_size [number]
--port [number]
--db_version [number]
--fork [num_forks]
```

I don't think any of those will have `1` as an input.

modules/Bio/EnsEMBL/VEP/Config.pm
diegomscoelho: Yes. It should not have any conflict with those names.

modules/Bio/EnsEMBL/VEP/Config.pm
diegomscoelho: Yes, two flags:
```sh
--shift_hgvs
--clin_sig_allele
```
Refactored code to check `%DEFAULTS` and remove from full command any flag equals to default usage. What you think?

nuno-agostinho (approved):
Hey @diegomscoelho, I tested the changes and they work for `--shift_hgvs 0 --clin_sig_allele 0`. Seems good to merge, thanks!

nuno-agostinho:
Merged to `main` and `release-109`.

PR #1300, opened 2022-11-04
Improve INSTALL.pl

nuno-agostinho:
- [ENSVAR-1979](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1979): Avoid checking VEP updates when using `--AUTO`
- [ENSVAR-5235](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5235): Print warnings using `warn()` instead of `print()`
- [ENSVAR-5234](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5234): Include `use warnings;` and reorder code
- [ENSVAR-5188](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5188): Improve config arguments and promote consistency with VEP arguments
    - Prefer lowercase arguments (uppercase still supported)
    - Added new alternative arguments that are more consistent with VEP (e.g., `--dir_cache` and `--dir_plugins` are now supported in `INSTALL.pl`)
    - Also added (alternative) arguments by adding an underscore to separate two words (e.g., `--cache_url` and `--fasta_url `)
    - Eventually, we will need to update documentation for INSTALL.pl (not critical as old usage is still compatible)

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -36,6 +36,8 @@ INSTALL.pl - a script to install required code and data for VEP
 =cut
 
 use strict;
+use warnings;
+
 use FindBin qw($RealBin);
 use lib $RealBin.'/modules';
 use Getopt::Long;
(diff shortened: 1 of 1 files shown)


nuno-agostinho:
Given recent issues with `INSTALL.pl` script, these changes will be discarded for now.

PR #1313, opened 2022-11-24
Support if allele_string is indel in VEP default input

likhitha-surapaneni:
[ENSVAR-3378](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-3378) 

The idea is to support indels in VEP default input and to get a minimised allele

### Testing
`vep --shift_3prime -i input.txt`

### Input
3 46358466 46358467 AA/AAG 1
3 46358467 46358467 A/AG 1
3 46358468 46358467 -/G 1
3 46358466 46358466 A/C 1
3 46358470 46358471 TA/GTA 1

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
@@ -865,9 +865,11 @@ sub pick_VariationFeatureOverlapAllele_per_gene {
 sub VariationFeature_to_output_hash {
   my $self = shift;
   my $vf = shift;
+  my $uploaded_allele_string = $vf->{original_allele_string} || $vf->{allele_string};
 
   my $hash = {
     Uploaded_variation  => $vf->variation_name ne '.' ? $vf->variation_name : ($vf->{original_chr} || $vf->{chr}).'_'.$vf->{start}.'_'.($vf->{allele_string} || $vf->{class_SO_term}),
+    Original_allele     => $uploaded_allele_string,
     Location            => ($vf->{chr} || $vf->seq_region_name).':'.format_coords($vf->{start}, $vf->{end}),
   };
 
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
@@ -868,7 +868,7 @@ sub VariationFeature_to_output_hash {
 
   my $hash = {
     Uploaded_variation  => $vf->variation_name ne '.' ? $vf->variation_name : ($vf->{original_chr} || $vf->{chr}).'_'.$vf->{start}.'_'.($vf->{allele_string} || $vf->{class_SO_term}),
-    Original_allele     => $vf->{original_allele_string} || $vf->{allele_string};,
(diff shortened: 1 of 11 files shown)


dglemos (commented):
In the VCF output, the MINIMISED is not in the header. I think this is why the value 1 (MINIMISED=1 in txt format) is not included in the output.

dglemos:
To be able to display the original alleles, you could start looking at this method https://github.com/Ensembl/ensembl-vep/blob/release/110/modules/Bio/EnsEMBL/VEP/OutputFactory.pm#L1246 
or here https://github.com/Ensembl/ensembl-vep/blob/release/110/modules/Bio/EnsEMBL/VEP/OutputFactory.pm#L871

dglemos:
A lot of tests are failing due to 'Original_allele' in the output.

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -865,9 +865,11 @@ sub pick_VariationFeatureOverlapAllele_per_gene {
 sub VariationFeature_to_output_hash {
   my $self = shift;
   my $vf = shift;
+  my $uploaded_allele_string = $vf->{original_allele_string} || $vf->{allele_string};
```
nakib103, line 868: `$uploaded_allele_string` is not used elsewhere. a redundant declaration. It would be better to put the logic in L872 directly.

modules/Bio/EnsEMBL/VEP/Parser/VEP_input.pm
```diff
@@ -159,6 +159,9 @@ sub create_VariationFeatures {
 
   # normal vf
   else {
+    my $is_in_del = 0;
```
nakib103, line 162: `$is_indel` will be better for readability.

modules/Bio/EnsEMBL/VEP/Parser/VEP_input.pm
```diff
@@ -159,6 +159,9 @@ sub create_VariationFeatures {
...
+    $is_in_del = 1 unless $allele_string =~ /[ATGC]{n}\/[ATGC]{n}/ or $allele_string =~ /-/;
```
nakib103, line 163: I think, the idea here is to discard variant class other than `indel` which would be SNV, substitution, simple insertions and deletions.

`/-/` takes care of the simple insertions and deletions. `/[ATGC]{n}\/[ATGC]{n}/` supposed to take care of SNVs and substitutions . But the problem is, here `n` needs to be a literal integer value. So the regex is not doing anything here (see `3 46358466 46358466 A/C 1` example - it is setting `$is_in_del = 1`)

Unfortunately, we also do not know what this value should be in case of substitution. Splitting the string and making sure the two parts is equal seems the correct approach here.

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/Parser/VEP_input.pm
likhitha-surapaneni: Hi @nakib103 , the idea is to support insertions or/and deletions and only discard for SNPs or already minimised allele strings

nakib103 (approved):
Looks good now, thanks for the change!

Issue #1357, opened 2023-02-15
VEP 109 tests failed with custom installation path


Stikus:
## Describe the issue
Header checker don't work with custom installation path

### System
- VEP version: 109.2
- VEP Cache version: 109
- Perl version: 5.26.1
- OS: Ubuntu 18.04
- tabix installed

### Full VEP command line
```
cd "$SOFT/ensembl-vep-${VEP_VERSION}" &&     perl -e "use Bio::DB::HTS::Tabix" &&     perl -e "use Bio::DB::BigFile" &&     prove -v t/*.t
```

### Full error message
```
#15 107.8 Test Summary Report
#15 107.8 -------------------
#15 107.8 t/OutputFactory_Tab.t                            (Wstat: 256 Tests: 17 Failed: 1)
#15 107.8   Failed test:  8
#15 107.8   Non-zero exit status: 1
#15 107.8 t/OutputFactory_VCF.t                            (Wstat: 256 Tests: 56 Failed: 1)
#15 107.8   Failed test:  5
#15 107.8   Non-zero exit status: 1
#15 107.8 t/OutputFactory_VEP_output.t                     (Wstat: 256 Tests: 20 Failed: 1)
#15 107.8   Failed test:  10
#15 107.8   Non-zero exit status: 1
#15 107.8 Files=49, Tests=1889, 107 wallclock secs ( 0.17 usr  0.07 sys + 102.92 cusr  4.30 csys = 107.46 CPU)
#15 107.8 Result: FAIL
```

```
#15 69.96 #   Failed test 'headers - plugin'
#15 69.96 #   at t/OutputFactory_Tab.t line 121.
#15 69.96 #                   '## VEP command-line: vep --assembly GRCh38 --cache_version 84 --database 0 --dir [PATH]/ensembl-vep-109.2[PATH]/ --input_file [PATH]/ensembl-vep-109.2[PATH]/test.vcf --no_stats --offline --plugin TestPlugin --quiet --show_ref_allele --tab --version 109.2#Uploaded_variation	Location	Allele	Gene	Feature	Feature_type	Consequence	cDNA_position	CDS_position	Protein_position	Amino_acids	Codons	Existing_variation	REF_ALLELE	IMPACT	DISTANCE	STRAND	FLAGS	test'
[...]

PR #1359, opened 2023-02-15
Hotfix: ignore invalid params from ENV vars and INI config file

nuno-agostinho:
https://github.com/Ensembl/ensembl-vep/issues/1357 and related with https://github.com/Ensembl/ensembl-vep/pull/1358

One of the params defined via environment variables is being passed to VEP installation tests, even when it is from an invalid parameter. This PR allows to silently ignore invalid parameters passed from environment variables and INI config files. The validity of the params is based on GetOptions.

## Testing

The following command should ignore the parameter `--version` but include `--spdi` in `VEP command-line` in the header.

```
export VEP_VERSION=109.2
export VEP_SPDI=1
vep -i variants.vcf
```


t/testdata/fasta/Homo_sapiens.GRCh38.toplevel.test.fa.fai
```diff
@@ -0,0 +1 @@
+21	46709983	57	60	61
```
nakib103, line 1: Hi @nuno-agostinho , Is this file redundant?

nakib103 (changes requested):
One main issue still remains that even if someone uses a valid flag (e.g. `VEP_DIR`) the flag will be accepted (setting `config->{dir} = <some value>`) and disrupting the test and potentially a VEP run.

One solution I can think of is to -
- read ENV variable only when running from docker.
- have a flag to disable reading from ENV variable even in docker (to prevent potential unwanted reading of variable like `VEP_DIR`).

t/testdata/fasta/Homo_sapiens.GRCh38.toplevel.test.fa.fai
nuno-agostinho: Removed the file, I don't think we need this, was just for testing.

nuno-agostinho:
Hi @nakib103, if they use a valid environment flag, then I think that VEP should accept it in all situations. We fixed `VEP_VERSION` because it was using a deprecated argument in `INSTALL.pl` and a non-existing argument in `vep`.

> One main issue still remains that even if someone uses a valid flag (e.g. VEP_DIR) the flag will be accepted (setting config->{dir} = <some value>) and disrupting the test and potentially a VEP run.

The test will be disrupted if any extra flag is given, whether it comes from environment variables, config INI file (that is automatically read from the user folder) or the command line. The only way to fix that is by making the test more flexible and to accept passing extra arguments, but I don't think this is a priority.

> read ENV variable only when running from docker.

There is no way to confirm we are in a Docker image (and we also want this to work in Singularity, etc.). Moreover, we want advanced users to have access to this option if they want to use them in any environment.

> have a flag to disable reading from ENV variable even in docker (to prevent potential unwanted reading of variable like VEP_DIR).

To avoid unwanted reading of variables, users should simply rename their variables so they don't start with `VEP_`. It is usual for command-line programs to accept environment variables that are prefixed with the name of the program.

nakib103 (approved):
After discussion, we will update our doc in e110 to have a not on usage of ENV variable with VEP_ prefix. 
LOTM !~

PR #1366, opened 2023-03-02
Docker fix: silence warnings for invalid options from ENV and INI config

nuno-agostinho:
After #1361, when running VEP in Docker, we get lots of warnings regarding pre-set environment variables for `INSTALL.pl` that are invalid in VEP:

```bash
$ vep --id rs699 --database --db_version 109 --force
Unknown option: no_update
Ignoring unsupported option 'no_update' found via ENV variable or INI file
Unknown option: no_plugins
Ignoring unsupported option 'no_plugins' found via ENV variable or INI file
Unknown option: pluginsdir
Ignoring unsupported option 'pluginsdir' found via ENV variable or INI file
Unknown option: no_htslib
Ignoring unsupported option 'no_htslib' found via ENV variable or INI file
```

These flags are good to have set up at all times in Docker to abstract users from the internal directory of the Docker image. This PR silently discards invalid options from ENV and INI config file by using the option `pass_through` from GetOpts.

### Testing

Run the following code with and without the changes in this PR:
```bash
export VEP_VERSION="109.3"
export VEP_JSON=0
vep --id rs699 --database --db_version 109 --force
```

* VEP should not print warning messages with these changes.
* In the command line of the VCF output, `--version 109.3` does not appear (which means it was silently discarded, as desired).

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -59,6 +59,7 @@ package Bio::EnsEMBL::VEP::Config;
 
 use File::Spec;
 use Getopt::Long;
+Getopt::Long::Configure("pass_through");
 
 use Bio::EnsEMBL::Utils::Exception qw(throw warning);
 use Bio::EnsEMBL::Variation::Utils::VariationEffect;
@@ -981,11 +982,7 @@ sub is_valid_param {
 
   # Run GetOptions to check validity of parameter
   my $res = {};
-
-  # Ignore STDERR from GetOptions to avoid warnings about invalid params
-  GetOptions($res, @VEP_PARAMS) or
-    warn("Ignoring unsupported option '${key}' found via ENV variable or INI file\n");
-
+  GetOptions($res, @VEP_PARAMS);
   my $is_valid = %$res ? 1 : 0;
 
   # Restore command-line arguments
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -1042,7 +1042,7 @@ sub read_config_from_file {
 
   close CONFIG;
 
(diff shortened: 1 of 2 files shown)


nakib103 (changes requested):
Thanks, it is working as intended and removing all warnings. Should we still keep the warning we are giving using `warn`? Because VEP do try to read from ENV variable and it would be better if we have log of what read from there as it is generally hidden (unlike cli/ini params).

In that respect we should also print out the ones we were able to read from the ENV. What is your thought on this?

nuno-agostinho:
Hey @nakib103,

I think printing environment variables that are ignored just pollutes the log and may mislead users to think something is wrong (such as in the case above with the Docker image). How about if we only print the ignored environment variables with `--verbose`?

Regarding reading from an environment variable, this would have helped avoid some issues in #1357. I would say we could print something like:
```
Setting --database=1 from environment variable VEP_database
Setting --json=1 from environment variable VEP_json
```

Do you think this should this be `--verbose` only or printed at all times?

Thanks for your comments!

nakib103:
Yeah, both of them in `--verbose` mode seems logical. They are meant for debugging and I agree having them always pollutes the log.

nuno-agostinho:
Hey @nakib103, changed the information about reading environment variables to appear when using `--verbose` only.

You can test this by running:
```
export VEP_VERSION="109.3"
export VEP_JSON=0
export VEP_PLUGIN=Downstream
./vep --id rs699 --database --db_version 109 --force --verbose --plugin NMD
```

That should print something like:
```
2023-03-08 16:04:43 - Set 'json=0' from environment variable VEP_JSON
2023-03-08 16:04:43 - Ignored unsupported option 'version=109.3' from environment variable VEP_VERSION
2023-03-08 16:04:43 - Appended 'plugin=Downstream' from environment variable VEP_PLUGIN (full value of plugin: ['NMD', 'Downstream'])
2023-03-08 16:04:43 - Read configuration from environment variables
```

PR #1376, opened 2023-03-13
nextflow-vep: split by number of variant

nakib103:
[ENSVAR-5530](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5530)

- split input VCF further by number of variant to increase speed. Added `--bin_size` to indicate we are using this option and mention the number of variant we want in each file.
- fix issues running on larger file. `bcftools sort` creates temp dir for large file and it fails in codon (/tmp is not writable).
- add option `--skip_check` - enables skip checking for index file. It helps to use cache for the first module if `-resume` is used.
- remove bcftools header lines in the output file. We use `--no-version` command given by bcftools which suppress this header lines to be printed out.


nuno-agostinho (approved):
Everything seems to be working fine, thanks @nakib103! :)

PR #1389, opened 2023-03-28
Supporting multiple inputs

likhitha-surapaneni:
- Support multiple inputs
- Support relative paths as input
- Create modular workflow
- [Rename output files based on input filename](https://github.com/Ensembl/ensembl-vep/pull/1389/commits/480199fd819dcc5c90578d7f31c7a3aa45aa165a)


nuno-agostinho (approved):
Approved @likhitha-surapaneni's commits! I added some changes, what do you think?

nextflow/workflows/run_vep.nf
```diff
@@ -81,10 +59,44 @@ else
   exit 1, "Undefined --vep_config parameter. Please provide a VEP config file"
 }
 
+// Input required
+if( !params.vcf && !params.vcf_dir) {
```
nakib103, line 63: We can put this inside `workflow{}`. In this way we can call the `workflow vep{}` from other workflow by giving vcf file not through params.

nakib103 (changes requested):

nextflow/workflows/run_vep.nf
likhitha-surapaneni: Makes sense @nakib103 , will modify accordingly

nextflow/workflows/run_vep.nf
```diff
@@ -11,12 +11,17 @@ nextflow.enable.dsl=2
 params.cpus = 1
 params.vep_config = "$PWD/vep_config/vep.ini"
 params.outdir = "outdir"
+
+params.vcf = null
+params.vcf_dir = null
```
nuno-agostinho, line 16:
```suggestion
```

nextflow/workflows/run_vep.nf
```diff
@@ -81,10 +59,39 @@ else
...
+    vcfInput = file(vcf)
+    if( !vcfInput.exists() ) {
+      exit 1, "The specified VCF input does not exist: ${vcfInput}"
+    }
+
+    if (vcfInput.isDirectory()) {
+      checkVCF(Channel.fromPath("${vcfInput}/*.gz"))
+    } else {
+      checkVCF(vcfInput)
+    }
```
nakib103, line 82: One problem I faced is when I send channel as `vcf` it will error out in file function. One way I solved is to check if `vcf` is string, might be better way to do it - 

```suggestion
    if(vcf instanceof String){
      vcfInput = file(vcf)
      if( !vcfInput.exists() ) {
        exit 1, "The specified VCF input does not exist: ${vcfInput}"
      }

      if (vcfInput.isDirectory()) {
        checkVCF(Channel.fromPath("${vcfInput}/*.gz"))
      } else {
        checkVCF(vcfInput)
      }
    }
    else {
      checkVCF(vcf)
    }
```

nakib103 (changes requested):

nextflow/nf_modules/run_vep.nf
```diff
@@ -22,19 +22,18 @@ process runVEP {
       2) A tabix index for that VCF output file
   */
   publishDir "${params.outdir}/vep-summary",
-    pattern: "${prefix}-*.vcf.gz_summary.*",
+    pattern: "${original}-${prefix}-*.vcf.gz_summary.*",
     mode:'move'
   cpus params.cpus
```
nakib103, line 27: is removing the container here is intentional?

nakib103 (changes requested):

nextflow/nf_modules/run_vep.nf
nuno-agostinho: Yes, @nakib103, we are using labels (`label 'vep'` and `label 'bcftools'`) so it's easy to change the VEP/bcftools version used for all processes in [nextflow.config](https://github.com/likhitha-surapaneni/ensembl-vep/blob/32b01aadbcc9fabdc0ba1d8a12a40ad9ebc613ab/nextflow/nextflow.config#L23-L29)

likhitha-surapaneni (commented):
LGTM

nakib103 (approved):
Looks good!~ thanks for the great work guys.

PR #1390, opened 2023-03-29
Nextflow: Minor improvements

likhitha-surapaneni:
- Support non-bgzipped files
- Support index files if they exist in the same directory as their VCF files
- Removed sorting in mergeVCF ([ENSVAR-5574](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5574))
- Checks moved to inside the workflow

### DIFF
--- a/nextflow/workflows/run_vep.nf
+++ b/nextflow/workflows/run_vep.nf
@@ -46,22 +46,6 @@ Options:
   exit 1
 }
 
-// VEP config required
-if ( params.vep_config ){
-  vepFile = file(params.vep_config)
-  if( !vepFile.exists() ){
-    exit 1, "The specified VEP config does not exist: ${params.vep_config}"
-  }
-}
-else
-{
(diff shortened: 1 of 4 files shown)


nextflow/workflows/run_vep.nf
```diff
@@ -70,14 +54,31 @@ workflow vep {
...
+    else
+    {
+      exit 1, "Undefined --vep_config parameter. Please provide a VEP config file"
+    }
+
+    // Input required
+    if( !params.vcf) {
+      exit 1, "Undefined --vcf parameter. Please provide the path to a VCF file"
+    }
+
```
nakib103, line 73: It is better to put it inside the other workflow. This checks are specific to running this workflow directly. We would want to call this workflow from other pipeline without necessarily having `params.vcf`

nakib103 (changes requested):

nuno-agostinho (approved):
Looks good to me!

PR #1391, opened 2023-03-29
Nextflow: Support multiple VEP config files

nuno-agostinho:
- Support multiple VEP config files
- Raise error if setting multiple VCF files and VEP config files (currently unsupported)
- Allow to input a Channel of VCF and automatically add tabix index files
- Differentiate output filename to avoid name collisions
- Improve speed when checking bgzip and tabix files


nakib103 (approved):
Tested and it works fine, thanks Nuno!

PR #1398, opened 2023-04-18
Running nextflow-vep from another pipeline

nakib103:
created on top of - https://github.com/Ensembl/ensembl-vep/pull/1391; so should be merged after that PR.

- improving on https://github.com/Ensembl/ensembl-vep/pull/1391, separate process for processing vcf input.
- extended above feature to also include `output_dir` in that relationship. It will allow output for different vcf file to output in different directory.
- separated the `splitVCF` modules to `generateSplits` which generate the split files and `splitVCF` which actually do the splitting based on generated split files in previous steps. Enables faster run time.
- support `.csi` index

Minor updates:
- updated the name of the output file to be `<original_name>_VEP.vcf.gz`, before it was a bit weird - `<original_name>.vcf.gz_vep_out.vcf.gz`
- some refactoring on variable and input/output module parameters


nextflow/nf_modules/generate_splits.nf
```diff
@@ -0,0 +1,38 @@
...
+/* 
+ * Read chromosomes 
+ */
+
+nextflow.enable.dsl=2
+
+// defaults
+prefix = ""
+params.outdir = ""
+params.cpus = 1
```
nuno-agostinho, line 12: Avoid setting parameters on nf_modules.

```suggestion
prefix = ""
```

nextflow/nf_modules/merge_VCF.nf
```diff
@@ -7,46 +7,41 @@
...
+  sorted_vcfs=\$(echo ${vcf_files} | xargs -n1 | sort | xargs)
+  bcftools concat --no-version -a \${sorted_vcfs} -Oz -o ${merged_vcf}
+  
+  if [[ ${index_type} == "tbi" ]]; then
+    bcftools index -t ${merged_vcf}
+  else
+    bcftools index -c ${merged_vcf}  
+  fi
+  
+  mv ${merged_vcf}* ${output_dir}
```
nuno-agostinho, line 45: We need to create the output directory before moving the files there:

```suggestion
  mkdir ${output_dir}
  mv ${merged_vcf}* ${output_dir}
```

nextflow/workflows/run_vep.nf
```diff
@@ -35,71 +37,84 @@ Usage:
...
-    mergeVCF(runVEP.out.vcf.groupTuple())
+    
+    vcf = createChannels(vcf, pattern="*.{vcf,gz}", true)
+    vep_config = createChannels(vep_config, pattern="*.ini", false)
+    
+    vcf.count()
+      .combine( vep_config.count() )
+      .subscribe{ if ( it[0] != it[1] && it[0] != 1 && it[1] != 1 ) 
+        exit 1, "Cannot map VCF and VEP config files to one-to-one, many-to-one, or, one-to-many scenario" 
+      }
```
nuno-agostinho, line 93: This removes functionality available before. We want to be able to:
- process multiple VCF files with a single VEP config
- process a single VCF file with multiple VEP configs

nextflow/workflows/run_vep.nf
```diff
@@ -35,71 +37,84 @@ Usage:
...
+        
+    // process input and create Channel
+    // this works like 'merge' operator and thus might make the pipeline un-resumable
+    // we might think of using 'toSortedList' and generate appropriate input from the 'processInput' module
+    processInput(vcf, vep_config, output_dir)
```
nuno-agostinho, line 98: This does not work: Channels are asynchronous which means the mapping between `vcf` and `vep_config` may be random.

I suggest to have a process/Nextflow function to get `vcf` and `vep_config` before creating Channels, map the data and then return a Channel with the mapping.

nuno-agostinho (changes requested):
Hey @nakib103, thanks for the changes! I made a few comments, namely the most important ones are:
- VCF files and VEP configs cannot be mapped based on two independent channels. This may result in random mapping of files, for instance, when using resume function; Nextflow does not guarantee the Channel order.
- Previous functionality to run VEP using multiple VCF files with a single VEP config and a single VCF file using multiple VEP configs are not working.

nuno-agostinho:
* Merged to `postreleasefix/111`, `release/111` and `main`.
* Backported to `postreleasefix/110` and `release/110`.

PR #1406, opened 2023-04-26
Use `--max_sv_size` as the HGVSg region limit

nuno-agostinho:
[ENSVAR-2603](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2603)

* This PR sets the default region size limit for HGVSg to `--max_sv_size` instead of the hardcoded 5kb limit.
* This also means that the default region size limit for HGVSg has increased from 5kb to 10Mb.
* Note: a HGVSg with a deletion around 10Mb takes ~1m30s to process.
* This PR requires changes from https://github.com/Ensembl/ensembl-variation/pull/985.

**Testing**

1. You can use the following HGVSg identifiers to test the limit with `--max_sv_size`:

```
16:g.89820416_89883065del
16:g.89818265_89883065del
16:g.79883066_89883065del
16:g.79818265_89883065del
NC_012920.1:m.1_10000000del
```

2. For HGVSg cases where the region limit goes above the limit, a warning message is printed:

```
WARNING: Unable to parse HGVS notation '16:g.79818265_89883065del'
MSG: Region requested must be smaller than 10000000. This value can be adjusted with --max_sv_size.
```

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm
@@ -166,6 +166,7 @@ sub create_VariationFeatures {
   my $vfa = $self->get_adaptor('variation', 'VariationFeature');
   my $vfs = [];
   my @errors;
+  my $max_size = $self->param('max_sv_size') || 5000;
 
   foreach my $core_group(@core_groups) {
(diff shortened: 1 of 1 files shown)


nakib103 (approved):
Thanks working as expected.
We might want to re-visit the default 10Mb max_sv_size in future.

PR #1407, opened 2023-04-27
Integrate StructuralVariantOverlap plugin features

nuno-agostinho:
[ENSVAR-5028](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5028): improved SV matching for custom annotation

This PR intends to bring all options from StructuralVariantOverlap (SVO) plugin to the **custom** option.

New option in custom | Option in SVO plugin | Description
-- | -- | --
file | file | VCF file of reference data
overlap_cutoff | overlap | percentage overlap between SVs
reciprocal | reciprocal | boolean: calculate reciprocal overlap
fields | cols | list of data types to return from the INFO fields
same_type | same_type | Boolean: report SV of the same type (eg deletions for deletions)
distance | distance | Required distance to the ends of the overlapping SVs to be considered
type: **overlap**, **within**, **surrounding**, **exact** | match_type: **surrounding**, **exact** | Only report reference SV which lie within (**within**) or completely surround (**surrounding**) the input SV options
short_name | label | annotation label (prefix) that will appear in the output

## Associated PR

- https://github.com/Ensembl/public-plugins/pull/700

## Implementation notes

- All options added apply to any **custom** file format: VCF, GTF, GFF, etc.
- **`overlap_cutoff`** default is `80` in SVO plugin and `0` in `--custom`
    - I left `0` as the default, but would it make more sense to use `80`?
- `match_type` from SVO plugin was integrated into `type`
- The subroutine `get_SO_term` comes from https://github.com/Ensembl/ensembl-vep/pull/1370

## Testing

Test the new flags and see if their results make sense:

```
--custom file=input/gnomad_v2.1_sv.sites.vcf.gz,short_name=test,fields=PC%info_columns,format=vcf,type=within,reciprocal=1,distance=1000,overlap_cutoff=80
```


likhitha-surapaneni (approved):
Looks good, thanks @nuno-agostinho

nuno-agostinho:
- Merged into `postreleasefix/111`, `release/111`, `main`.
- Backported to `postreleasefix/110` and `release/110`.

Issue #1405, opened 2023-04-26
--custom gnomAD genomes annotation does not return variant in overlap mode


FlorianErger:
Dear VEP team,

running vep with the --custom flag pointing to the gnomAD genomes data in overlap mode does not return the allele frequency of a variant at that exact position. It does return the variant in exact mode. I have a local setup of VEP (local cache, local database). I have downloaded gnomad.genomes.r2.0.1.sites.noVEP.vcf.gz for custom annotation as described in the VEP documentation. I slightly modified this file to have an info field with the reference sequence, variant sequence and position for easier downstream data management, but do not believe this has any bearing on my issue, because 99.9% of variants are annotated normally using these files.

A variant where my issue occurs is "3:g186302392G>T" [hg19]. If I

`zcat gnomad.exomes.r2.0.1.sites.noVEP.vcf.gz | grep 186302392`

I thus get
[...]

PR #1410, opened 2023-05-03
bugfix: `overlap` in --custom option with VCF file

nakib103:
Reported in https://github.com/Ensembl/ensembl-vep/issues/1405
JIRA - https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5628

The issue is twofold - 
- for multi-allelic variant with SNV and indel we are only considering indel (see [here](https://github.com/Ensembl/ensembl-io/blob/4946a8655e8fdbcf37654c80144a9b09e97604b6/modules/Bio/EnsEMBL/IO/Parser/BaseVCF4.pm#L269)). If one allele is indel we are updating start position by 1 which would be wrong of any of the allele is SNV.
- for deletion we are getting end co-ordinates with 1 extra bp (see [here](https://github.com/Ensembl/ensembl-io/blob/4946a8655e8fdbcf37654c80144a9b09e97604b6/modules/Bio/EnsEMBL/IO/Parser/BaseVCF4.pm#L322)). As for indel start is already incremented by 1 - adding `len(ref) - 1` actually gives us 1 extra bp length. We should add `len(ref) - 1` to raw_start instead.

But can be solved just by updating the end co-ordinates described in the second point. Added a fix in VEP parser instead of ensembl-io because that affect can be wide spread.

### Testing
custom VCF file -
```
3	186302389	.	TTA	T	.	.	CLNDISDB=test_1
3	186302392	.	G	T,GTGTGTGTGTGTGTGTGTGTA	.	.	CLNDISDB=test_2
```

Command:
```
vep --id "3:g.186302392G>T" --database --db_version 108 --force --assembly GRCh37  --custom file=<file>,short_name=test,format=vcf,type=overlap,coords=0,fields=CLNDISDB
```

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm
@@ -70,6 +70,7 @@ package Bio::EnsEMBL::VEP::AnnotationSource::File::VCF;
 
 use Bio::EnsEMBL::Utils::Exception qw(throw warning);
 use Bio::EnsEMBL::Utils::Sequence qw(reverse_comp);
+use Bio::EnsEMBL::Variation::Utils::VariationEffect qw(overlap);
 use Bio::EnsEMBL::Variation::Utils::Sequence qw(get_matched_variant_alleles);
 use Bio::EnsEMBL::IO::Parser::VCF4Tabix;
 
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm
```diff
@@ -324,26 +325,45 @@ sub _record_overlaps_VF {
...
-    },
-    {
-      ref  => $parser->get_reference,
-      alts => $parser->get_alternatives,
-      pos  => $parser->get_raw_start,
-    }
-  );
+  if ($type eq 'overlap') {
+    my $parser_start = $parser->get_raw_start;
+    my $parser_end = $parser_start + length($parser->get_raw_reference) - 1;
```
nuno-agostinho, line 336: Are these two lines the only changes to `_record_overlaps_VF`? Why not fix in that function instead of duplicating code? This fix as is is in conflict with https://github.com/Ensembl/ensembl-vep/pull/1407, but if we fixed in the function, no problem would arise.

nuno-agostinho (changes requested):
Hi @nakib103, thanks for the fix. I have some comments about this PR.

travisci/harness.sh
```diff
@@ -20,7 +20,7 @@ fi
 
 rt=$?
 if [ $rt -eq 0 ]; then
-  if [ "$COVERALLS" = 'true' ]; then
+  if [[ "$COVERALLS" = 'true' && ! "${TRAVIS_PERL_VERSION}" =~ 5.10 ]]; then
```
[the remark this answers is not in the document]
nakib103: This change was made to fix travis build error for coverage. These changes are now already merged through separate PR. I have reverted the change here.

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm
nakib103: The other function (I assume you are talking about `SUPER::_record_overlaps_VF(@_)`) is a generic function for any parser, so we cannot add this logic here which depends on specifically VCF parser.

I saw you delegated all the new `types` in custom to the `SUPER::_record_overlaps_VF(@_)` function. You need to account that `parser->get_end` can give you one extra bp.

It would be best if we can put this logic in `$parser->get_end` but we need to assess the impact of changing this upstream of a function.

nuno-agostinho (approved):

PR #1411, opened 2023-05-04
Stop coverage for perl 5.10

nakib103:
Devel::Cover have updated there minimal Perl supported version to v5.12.0 -
https://github.com/pjcj/Devel--Cover/blob/v1.40/Changes (see 1.39 release)

All ensembl-vep travis build will fails on 5.10 coverage test.

- temporarily stopping coverage for perl 5.10
- update trigger-dependency stage to use 5.14 (it was using perl 5.10 before)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ jobs:
   include:
     - stage: trigger_dependent_builds
       script: ./travisci/trigger-dependent-build.sh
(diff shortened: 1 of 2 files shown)


likhitha-surapaneni (approved):
LGTM! Tests not failing anymore

PR #1413, opened 2023-05-11
Print only file name as gff/gtf source

nakib103:
With https://github.com/Ensembl/ensembl-vep/pull/1317 the `source_name` has been default to `file`. It results in full file path being printed in the output. before the logic was to print only the file name in the output.

Revert back to previous logic.

### Test
```
chr2	32198	.	A	G
```

```
vep  --i  <input_file> --assembly GRCh37 --fasta path/to/CHM13.fa --gff path/to/Homo_sapiens-GCA_009914755.4-2022_06-genes.gff3.gz --force --vcf
```

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSourceAdaptor.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSourceAdaptor.pm
@@ -229,7 +229,7 @@ sub get_all_custom {
     my $opts = {
       config => $self->config,
       file => $hash{"file"},
-      short_name => $hash{"short_name"} || $hash{"file"},
+      short_name => $hash{"short_name"},
       format => $hash{"format"},
       type => $hash{"type"} || "overlap",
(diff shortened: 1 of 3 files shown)


Gurpreet-Ghattaoraya:
> Output VCF supresses gff file paths in most occurances but it is still present on one line. e.g. of output run
> 
> ##INFO=<ID=Homo_sapiens.GRCh38.106.gff3.gz,Number=.,Type=String,Description="`/directory/path/still/present/`Homo_sapiens.GRCh38.106.gff3.gz">
> 
> This will also need changing to supress here as well.
> 
> The filepaths are supressed in all other locations of the VCF file.

Comment has been addressed.
The path in the comment above is now masked as intended. Along with all other instances of reference of the GFF file.

Thanks @nakib103 !!

At time of this review - failing Travis check - related to a different pull request. @jamie-m-a is OK to overrule Travis and fine to merge.

Gurpreet-Ghattaoraya (approved):

PR #1421, opened 2023-05-24
Add option for HGVSp prediction format

nakib103:
related to - https://github.com/Ensembl/ensembl-variation/pull/994


nuno-agostinho (approved):
Thanks @nakib103!

PR #1423, opened 2023-05-25
Clinical significance - compare ref alleles 

dglemos:
[ENSVAR-5773](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5773)

Update VEP to compare the reference alleles before returning the clinical significance.
This update requires:  https://github.com/Ensembl/ensembl-variation/pull/995


nakib103 (approved):
Looks good, working as intended!

PR #1427, opened 2023-06-02
Fix results when enabling `--no_stats`

nuno-agostinho:
[ENSVAR-4494](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4494): related with #1034 

When running `--no_stats`, consequences for `--hgvs` can be wrongly calculated. This PR adds missing logic required when running `--no_stats` to return the expected results.

However, this adds extra steps that can make VEP slower with `-no_stats` -- we need to benchmark how slow adding this can be for a large amount of variants.

## Testing

Running with or without `--no_stats --hgvs` should return the same results when using the following input VCF:

```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT
chr3	42210085	.	C	CGGAGGA,CGGA	359	PASS	.
chr14	24300643	.	A	AGAGGAG,AGAG	694	PASS	.
```

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory.pm
@@ -952,7 +952,16 @@ sub VariationFeature_to_output_hash {
   # check_ref tests
   $hash->{CHECK_REF} = 'failed' if defined($vf->{check_ref_failed});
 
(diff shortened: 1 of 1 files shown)


serge2016:
Hello! Hope to see update here very much!

PR #1429, opened 2023-06-05
bugfix: Refseq MT transcripts are not annotated from cache

nakib103:
MT transcripts are not being annotated by VEP when used with refseq cache with version 106 and above. The issue lies with change in transcript stable id for refseq cache from this [PR](https://github.com/Ensembl/ensembl-analysis/commit/954fcdc72a10049de78925dd748d412420e2162d#diff-9946f013146909c64697ed2002d30e863111c7fe0b68ceecbcfd0b4a5fd94479).

Details in - [ENSVAR-5791](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5791)

### Unit test
1. Unit test added to check that `filter_transcript` passes when using refseq cache and transcript has the 4 types of stable id that is currently supported - 
- NR_046198.3
- 4558
- rna-TRNK
- COX3

_**Needed to add some cache data files but they are small in size.**_

2. Also updated previous test case to check `filter_transcript` for cache type `merged`. We were not setting the cache type before - https://github.com/Ensembl/ensembl-vep/blob/2360ea1e337b337ff5fb33daa3f0c36e326b52cb/t/AnnotationSource_Cache_Transcript.t#L150

### Test

data - 
```
chrM	4745	.	A	G
chrM	4769	.	A	G
chrM	7337	.	G	A
```

command -
```
vep --force --i data_file --species homo_sapiens --assembly GRCh38 --cache $cache_dir --offline --fasta $fasta_file --cache_version 106 --refseq
```

Check with cache from before and after e106. It is best to also check with `merged` cache.


nuno-agostinho (approved):
Working as expected! Thanks @nakib103

nuno-agostinho:
Merged to `main` and `release/111`.

PR #1448, opened 2023-06-29
bugfix: resolve many-to-one situation in nextflow-vep

nakib103:
nextflow-vep many-to-one (and one-to-many) situation with VCF and VEP config file was not working.

This is because giving them as separate channel to `processInputs()` does not work as expected (it is supposed to work like `combine()`). Nextflow handles if one of them is not a channel. 

But going through that route was only considered because of many-to-many situation, and as we are ot supporting that situation I simply used `combine()` which works fine.

**Note** - also stopped channeling `output_dir` through the workflow, it is used only by the `mergeVCF` step. So we can just give input to that step.

### DIFF
--- a/nextflow/nf_modules/merge_VCF.nf
+++ b/nextflow/nf_modules/merge_VCF.nf
@@ -22,14 +22,19 @@ process mergeVCF {
   cache 'lenient'
    
   input:
-  tuple val(original_vcf), path(vcf_files), path(index_files), path(vep_config), val(index_type)
-  val output_dir
+  tuple val(original_vcf), path(vcf_files), path(index_files), val(vep_config), val(index_type),
+    val(one_to_many),
+    val(output_dir)
   
   output:
   val("${output_dir}/${merged_vcf}")
 
   script:
   merged_vcf = merged_vcf ?: file(original_vcf).getName().replace(".vcf", "_VEP.vcf")
+  merged_vcf = one_to_many ? merged_vcf.replace(
+    "_VEP.vcf", 
+    "_" + file(vep_config).getName().replace(".ini", "") + "_VEP.vcf"
+  ) : merged_vcf
   index_flag = index_type == "tbi" ? "-t" : "-c"
   
   """
--- a/nextflow/nf_modules/merge_VCF.nf
+++ b/nextflow/nf_modules/merge_VCF.nf
@@ -22,7 +22,7 @@ process mergeVCF {
   cache 'lenient'
    
   input:
-  tuple val(original_vcf), path(vcf_files), path(index_files), val(index_type)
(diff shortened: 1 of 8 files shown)


nextflow/nf_modules/run_vep.nf
```diff
@@ -22,10 +22,10 @@ process runVEP {
   label 'vep'
 
   input:
-  tuple val(original_vcf), path(vcf), path(vcf_index), path(vep_config), val(output_dir), val(index_type)
+  tuple val(original_vcf), path(vcf), path(vcf_index), path(vep_config), val(index_type)
   
   output:
-  tuple val(original_vcf), path(out_vcf), path("${out_vcf}.{tbi,csi}"), val(output_dir), val(index_type), emit: files
+  tuple val(original_vcf), path(out_vcf), path("${out_vcf}.{tbi,csi}"), val(index_type), emit: files
```
nuno-agostinho, line 28:
```suggestion
  tuple val(original_vcf), path(out_vcf), path("${out_vcf}.{tbi,csi}"), path(vep_config), val(index_type), emit: files
```

nextflow/nf_modules/merge_VCF.nf
```diff
@@ -22,7 +22,8 @@ process mergeVCF {
   cache 'lenient'
    
   input:
-  tuple val(original_vcf), path(vcf_files), path(index_files), val(output_dir), val(index_type)
+  tuple val(original_vcf), path(vcf_files), path(index_files), val(index_type)
```
nuno-agostinho, line 25:
```suggestion
  tuple val(original_vcf), path(vcf_files), path(index_files), path(vep_config), val(index_type)
```

nextflow/nf_modules/merge_VCF.nf
nuno-agostinho, line 1: In line 32, we need to add the name of the VEP config file to distinguish between the results:

```nextflow
  merged_vcf = merged_vcf ?: file(original_vcf).name.replace(".vcf", "_" + file(vep_config).simpleName + "_VEP.vcf")
```

nuno-agostinho (changes requested):
Hi Nakib, thanks for the PR! I have a comment:

If we use multiple INI config files for a single VCF, we will only have a single VCF file in the end. I put some suggestions on what you can change to avoid this.

See if you agree. Thanks!

nakib103:
Thanks Nuno! I have added the suggestion and fixed one-to-many situation.

nuno-agostinho (approved):
All is looking fine, @nakib103. Thanks!

nuno-agostinho:
* Merged to `release/111` and `main`
* Backported to `postreleasefix/110` and `release/110`

Issue #1433, opened 2023-06-14
Deadsnakes no longer supports Ubuntu 18.04


muppetjones:
### Context

- Extending the base docker image with newer versions of python (specifically 3.9)
- Compiling python from source results in very slow execution at runtime (presumably due to compiler flags, etc)
- Using deadsnakes PPA significantly improves speed and stability

### Issues 

- [Deadsnakes no longes supports python 3.9](https://github.com/deadsnakes/issues/issues/251)

### System

> NOTE: This same issue affects all released docker images

- VEP version: `ensemblorg/ensembl-vep:release_108.2`
- VEP Cache version: n/a
- Perl version: n/z
- OS: Ubuntu

### Full VEP command line

```dockerfile
ARG PYTHON_VERSION=3.9

RUN echo "=== Installing python ${PYTHON_VERSION}" \
    && add-apt-repository ppa:deadsnakes/ppa \
    && apt-get update && apt-get -yqq install \
    python${PYTHON_VERSION}-dev \
    python${PYTHON_VERSION}-venv \
    && echo "=== Installing pip for python version" \
    && curl -sS https://bootstrap.pypa.io/get-pip.py | python${PYTHON_VERSION} \
    && echo "=== Setting up v${PYTHON_VERSION} as the default python" \
    && ln -s /usr/bin/python${PYTHON_VERSION} /usr/local/bin/python3 \
    && ln -s /usr/local/bin/python3 /usr/local/bin/python \
    && echo "=== Cleanup apt-get" \
    && apt-get clean autoclean \
    && apt-get autoremove -y \
    && rm -rf /var/lib/apt/lists/*
```

### Full error message

```
#10 10.93 E: Unable to locate package python3.9-dev
#10 10.93 E: Couldn't find any package by glob 'python3.9-dev'
#10 10.93 E: Couldn't find any package by regex 'python3.9-dev'
#10 10.93 E: Unable to locate package python3.9-venv
#10 10.93 E: Couldn't find any package by glob 'python3.9-venv'
#10 10.93 E: Couldn't find any package by regex 'python3.9-venv'
```

### Data files (if applicable)
n/a

nakib103:
Hello @muppetjones ,

Thanks for flagging this to us. We will see the feasibility of updating Ubuntu to next LTS.

Best regards,
Nakib

nakib103:
Hello @muppetjones ,

[...]

PR #1464, opened 2023-07-24
Docker: update ubuntu version latest LTS (22.04)

nakib103:
[ENSVAR-5837](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5837)

Reported in - https://github.com/Ensembl/ensembl-vep/issues/1433

Ubuntu 18.04 have come to end-of standard support (see [here](https://wiki.ubuntu.com/Releases)). We are going to update docker image to latest LTS version which is 22.04 and have standard support until April 2027.

- update dependencies from python to python2 (https://github.com/Ensembl/VEP_plugins/pull/620)
- update docs and plugin usage to python2 command
- we also create symlink for python to python2 just in case.


likhitha-surapaneni (approved):
LGTM, thanks @nakib103

Issue #1023, opened 2021-07-12
Feature Request: Summary statistics option for custom annotations (bed/bigwig) as CNVs annotation can be enormous


davmlaw:
For example, a <1k CNV VCF (from Dragen) using custom annotations with VEP can produce a 1GB gzipped annotated VCF

This is because CNVs span long distances, and custom bed/bigwig annotations work by packing the values together in the fields, eg:

```
0.0570000000298023&0.372000008821487&0.638999998569489&0.579999983310699
```

Eg a single VCF record with REFLEN=15015  had per-base annotation from phastCons100way_vertebrate - so a single field was 257,903 characters long.

```
csq = v.INFO.get("CSQ")
td = dict(zip(vep_columns, transcript_csq.split("|")))
cons_data = td["phastCons100way_vertebrate"]
In [62]: len(cons_data)                                                                                                                      
Out[62]: 257903
```

**Suggested solution:** 

Have an option for custom annotations (bed/bigwig) to produce summary statistics, eg something like this:

```
In [65]: pd.Series(cons_data.split("&"), dtype='f').describe()                                                                                          
Out[65]: 
count    15015.000000
mean         0.066185
std          0.141911
min          0.000000
25%          0.002000
50%          0.013000
75%          0.058000
max          0.997000
dtype: float64
```

NavenM:
Hi @davmlaw ,

that sounds like a sensible solution, that we could potentially implement. I've added it to our backlog, but it might be a few months before we get to develop this.

Thanks

davmlaw:
Hi, in VEP10 it looks like you added new ```--custom``` named arguments, ie ```format```, ```fields``` etc.

For conservation - I only really need the max, so perhaps this could be implemented as adding a new field:

summary_stat=max

Where if you have >1 result (ie variant spans multiple bases) you apply that summary statistic on the data before returning it

summary_stat options could be eg one of "max", "mean", "median", "min"

If anyone wants multiple summary stats, they could just run custom multiple times with different values for summary_stat

davmlaw:
It's possible to calculate summary stats as you go (without keeping all of them in memory) - eg min/max is trivial, can do a rolling/moving mean - (though median is difficult)

If you did this rolling calculation, you would also likely solve #1414

nuno-agostinho:
Hi @davmlaw,

Thanks for your suggestions!

[...]

PR #1470, opened 2023-08-04
Custom annotations: summary statistics + filter by number of records

nuno-agostinho:
ENSVAR-5925, related with https://github.com/Ensembl/ensembl-vep/issues/1023

**UPDATE [20 Sept 2023]:** Calculating summary statistics has been extended to all annotation formats (as we can get scores from all formats, anyway). By default, summary statistics will be activated only for BED/BigWig files.

### Changelog

* [Extend all `--custom` options currently available to `bigwig` files](61527332c36cce32d887e5cf4e199a3268b6106a)
* **Add new `num_records` option to filter number of records appended to large variants**
  * Default: `50` records
  * `num_records=all` prints all records
  * If there are more records than `num_records`, the string `...` is appended after reaching the maximum number of records
* **Add new `summary_stats` option to summarise scores**
  * ~~Default: `min%mean%max%count%sum` for `bigwig`/`bed` files, `none` for others~~
      * **UPDATE c7f99e4c6b4ad2e467f8094d6820468edce38d1f:** other formats can also calculate statistics (if the user explicitly sets `summary_stats`)
      * **UPDATE 90a84b8:** default is now `none` for all formats to avoid issues with JSON output
  * Use `summary_stats=none` to avoid calculating any stats
  * Other statistics (`median`, `sd`, `q1`, `q3`) could be added, but require more time/memory
* [Throw error if using invalid `--custom` options](https://github.com/Ensembl/ensembl-vep/pull/1470/commits/b1e9f89d6fc9b235e721f1d2bff1a39cb3433541)

#### Notes

* If using `num_records=0`, no records are printed, only `...`
* ~~If `coords=1`, summary stats cannot be calculated (as we don't get the scores)~~
    * **UPDATE f3679ba014c4b742564083bbb365957aa8cf3110**: we now get scores regardless of the value of `coords`
* Using `num_records=50` (default) with `summary_stats=none` (default for non-biwig/bed files) should be faster and use less memory as it doesn't need to transverse the whole custom annotation anymore

### Public documentation

- https://github.com/Ensembl/public-plugins/pull/723

### Testing

* Test `num_records` option: `num_records=all`, `num_records=0`, `num_records=5` and default value
    - Test with a custom `VCF` to see if results are expected
* Test `summary_stats` option: available options are `min`, `mean`, `max`, `count`, `sum`; default is `min%mean%max%count%sum`

Command example:
```
perl vep \
     --i $vcf \
     --o output/vep-output.vcf --vcf \
     --offline \
     --allow_non_variant \
     --custom input/hg38.phyloP100way.bw,phyloP,bigwig,overlap,0
```

bigwig file example: [hg38.phyloP100way.bw](https://downloads.molgeniscloud.org/downloads/vip/resources/GRCh38/hg38.phyloP100way.bw)


dglemos:
The update looks good, there is only one small problem with json output.
If I run the default custom for bed, the JSON output format is missing the stats (min,mean,max,count,sum)

`--custom file=../custom_files/blood.funmotifs.bed.gz,short_name=my_bed,format=bed,coords=0 \`

Output:
`{"my_bed":[{"name":"ETV2_MA0762.1","score":12.0328}]}`

nuno-agostinho:
Hey @dglemos, I updated the JSON format to show summary statistics as expected. The default is now 'none' for all formats to avoid output issues in JSON.

I will also update this info in the docs: Ensembl/public-plugins#723.

dglemos:
Merged into release/111 and main

PR #1477, opened 2023-08-15
Fix web VEP input format detection

nuno-agostinho:
Standardise input format detection between web VEP and command-line VEP


nakib103 (approved):
Thanks Nuno! LOTM!~

PR #1478, opened 2023-08-17
Run VEP with SV IDs

nuno-agostinho:
ENSVAR-5793

Currently, VEP only returns non-structural variants when using variation IDs.

This PR expands the feature to also allow VEP to return SV using their ID[^1].

### Testing

```
perl vep --id esv1815690  --database --force
perl vep --id esv1850194  --database --force
perl vep --id nsv1000164  --database --force
perl vep --id cnvi0117333 --database --force
perl vep --id CN_571740   --database --force
```

This also works by running `perl vep -i input.txt --database --force`, where the input file contains the identifiers:
```
esv1815690
esv1850194
rs699
nsv1000164
cnvi0117333
CN_571740
```

[^1]: Some SV IDs supported in the Ensembl browser (like [`nsv443016`](http://wp-np2-11.ebi.ac.uk:6070/Sus_scrofa/StructuralVariation/Mappings?r=7:67148496-67211415;sv=nsv443016;svf=665;vdb=variation)) don't work because they are not part of the species database.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/BaseVEP.pm
+++ b/modules/Bio/EnsEMBL/VEP/BaseVEP.pm
@@ -792,7 +792,8 @@ sub skipped_variant_msg {
     $msg = "(" . $line . "): " . $msg;
   }
 
-  $msg = "Line $line_number skipped " . $msg if defined $line_number;
+  $msg = (defined $line_number and not defined $self->param('input_data')) ?
+    "line $line_number skipped " . $msg : "variant skipped " . $msg;
   $self->warning_msg("WARNING: " . $msg . "\n");
   $self->{skip_line} = 1;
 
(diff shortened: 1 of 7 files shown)


dglemos:
For consistency, shouldn't this PR also be for release 112?

nuno-agostinho:
Hey @dglemos, they are related, but this is a stand-alone and can go already in release 111. Do you think it would be better to delay it for 112?

dglemos:
That's true this one supports IDs for VEP.
The only implication is that in REST the VEP endpoint is going to work with SV IDs but variation is not.
```
vep/human/id/esv1815690?content-type=application/json -> will work
/variation/human/esv1815690?content-type=application/json -> won't work
```

dglemos (approved):

dglemos:
The code looks good.
I'll wait until you add a max size.

nuno-agostinho:
Hey @dglemos, added a new commit to generalise `max_sv_size` limit to all input formats. Before, it was only being applied to VCFs. Also, improved the error message to avoid showing "Line 1 skipped" when using `--id`.

dglemos:
In the public docs it could mention SV IDs are also valid: https://www.ensembl.org/info/docs/tools/vep/vep_formats.html#id

dglemos:
Merged into release/111 and main (conflicts resolved)

PR #1487, opened 2023-08-25
Clinical significance - remove clinvar ref allele check 

dglemos:
Revert https://github.com/Ensembl/ensembl-vep/pull/1423


jamie-m-a (approved):
Reverting to original code.

Issue #1476, opened 2023-08-14
Format for breakend variants in VCF


janrehker:
Hi,

According to https://github.com/Ensembl/ensembl-vep/issues/441 vep should now be able to handle breakend variants.

I tried 

1 234919885 bnd A [chr1:17124942[A . . SVTYPE=BND .

which is actually the example for vcf input in the structural variant section from https://www.ensembl.org/info/docs/tools/vep/vep_formats.html 
I tried to submit this variant on the vep web interface (vep v110) but unfortunately it did not work.

I might be holding it wrong.

Best regards,
Jan

nuno-agostinho:
Hi @janrehker,

Thanks for reporting this issue.

VEP indeed supports breakend variants (including that example), but for some reason it is not accepting it if you submit as the first variant. For now, you need to send another variant as the first.

We recently updated VEP to avoid this issue, which I just tested and is working for the command-line version, but not for the web interface. I am going to try to understand what is going on and I will let you know.

Kind regards,
Nuno

janrehker:
Thanks, for letting me know! Putting an snv in the first line worked for me indeed and our slightly modified Manta-results we used as a positive control were annotated as truncated, as we expected.

We're trying the command line version now.

Thanks a bunch for this new feature!


Best regards,
Jan

janrehker:
Hi,

I encounter a similar issue like @osowiecki in https://github.com/Ensembl/ensembl-vep/issues/441

In my case it is DELLY output:
WARNING: Line 187 skipped (2 227032111 BND00000186 A [1:106864407[A . Low...): start > end+1 : (START=227032112, END=106864407)

This is the line, as it went into vep:

2 227032111 BND00000186 A [1:106864407[A . LowQual IMPRECISE;SVTYPE=BND;SVMETHOD=EMBL.DELLYv0.7.8;CHR2=1;END=106864407;PE=2;MAPQ=60;CT=5to5;CIPOS=-231,231;CIEND=-231,231 GT:GL:GQ:FT:RCL:RC:RCR:CN:DR:DV:RR:RV 0/0:0,-19.006,-594:10000:PASS:8:5259:1:1169:101:2:0:0

This is the command line I used in vep 110.1:
[...]

PR #1399, opened 2023-04-18
Add support for breakend variants from VCF

nuno-agostinho:
[ENSVAR-2086](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2086), fixes #437, #441 and #782
* Documentation on breakends (BND) available in [VCF v4.2 Specification, chapter 5.4](https://samtools.github.io/hts-specs/VCFv4.2.pdf)
* Requires consequence updates in [ensembl-variation#984](https://github.com/Ensembl/ensembl-variation/pull/984)

## Logic
1. When creating structural variants from VCF, [parse breakend variants in VEP](https://github.com/nuno-agostinho/ensembl-vep/blob/47a8edb1a2fd42214ef4bfcdf5ee658ecdfd251d/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm#L528-L532)
        - Variants are considered breakend if they have `SVTYPE=BND` in INFO field or `<BND>` in ALT field, for instance:
```
chr1    234919885  MantaBND:1445:0:1:1:3:0:1     A  [chr1:17124942[A
chr22    17577704  ref_panel_1kg_v1_BND_chr22_6  N  <BND>             487 BOTHSIDES_SUPPORT;UNRESOLVED    ALGORITHMS=manta;CHR2=chr22;END=17577704;END2=20098034;EVIDENCE=PE,SR;PREDICTED_INTRONIC=DGCR8,SLC25A18;PREDICTED_NONCODING_BREAKPOINT=Tommerup_TADanno;STRANDS=++;SVLEN=2520330;SVTYPE=BND;UNRESOLVED_TYPE=INVERSION_SINGLE_ENDER_++
```
2. For each alternative allele (e.g., `A[chr22:22893780[` in ALT field or `CHR2=chr22;END2=20098034` in INFO field), [create a hashref with chromosome, position, placement (before/after) and if mate is inverted](https://github.com/nuno-agostinho/ensembl-vep/blob/47a8edb1a2fd42214ef4bfcdf5ee658ecdfd251d/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm#L423-L433)
    - This PR also supports multiple comma-separated alternative alleles in ALT field, such as `A[chr22:22893780[,CT[chrX:10932343[`
4. As usual, VEP creates an input buffer with a certain number of variants loaded into memory. For all variants in the input buffer:
[...]

PR #1491, opened 2023-08-30
Parse END field in Manta breakend variants

nuno-agostinho:
Fixes second issue mentioned in #1476: `END` cannot be a smaller number than variant start, according to [VCF 4.4 specifications](https://samtools.github.io/hts-specs/VCFv4.4.pdf), even though this is what happens in the output of breakends from the SV caller Illumina Manta.

As this is a popular SV caller, we will correctly parse the Manta breakend variant (even if out of VCF spec).

### Changelog

* Delete `END` and parse the variant end again (this way, variant end is the same as start)
* Store value of `END` as `END2`
    - If the alternative allele of breakend is similar to `[1:106864407[A`, the information in the INFO field is ignored
    - Otherwise, if both `CHR2` and `END2` are defined, they will be used to prepare the breakend information (logic already introduced in #1399)
* Remove unnecessary code regarding `END` that is already run in `$parser->get_end()`

### Testing

Run VEP with some breakend (and other) variants, such as:

```2 227032111 BND00000186 A [1:106864407[A . LowQual IMPRECISE;SVTYPE=BND;SVMETHOD=EMBL.DELLYv0.7.8;CHR2=1;END=106864407;PE=2;MAPQ=60;CT=5to5;CIPOS=-231,231;CIEND=-231,231 GT:GL:GQ:FT:RCL:RC:RCR:CN:DR:DV:RR:RV 0/0:0,-19.006,-594:10000:PASS:8:5259:1:1169:101:2:0:0```

I added unit tests to check changes.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
@@ -518,18 +518,19 @@ sub create_StructuralVariationFeatures {
   my $type = $info->{SVTYPE} || $alt;
   my $so_term = $self->get_SO_term($type) || $type;
 
+  ## Illumina Manta (SV caller) may use INFO/END to identify the position of the
+  ## breakend mate (unsupported based on VCF 4.4 specifications)
+  if ($so_term =~ /break/ && $parser->get_info->{END}) {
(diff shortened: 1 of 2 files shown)


nakib103 (approved):
Looks Good and work as expected, thanks @nuno-agostinho

PR #1497, opened 2023-09-14
nextflow-vep: separate out CLI params configuration 

nakib103:
nextflow-vep does not allow many-to-many scenario between input VCF and VEP config files. This PR allows for the case  where we use nextflow-vep from another pipeline. This is achieved by separating out the processing of CLI params and allocating that jobs to main workflow (which is used when we call nextflow-vep from CLI).

It is the responsibility of the workflow that is calling nextflow-vep to pass the VCF file and related VEP config inside a tuple.

**meta map**: we also introduce meta map as input parameter to vep workflow -
```
tuple val(meta), path(vcf), path(vcf_index), path(vep_config)
```
It is useful to use it instead of long length of parameters list in the process input. Some of which are only passed through to be used by some other downstream process.

### Test
You can create a test nextflow pipeline to test the new updates -
```
include { vep } from "${params.repo_dir}/ensembl-vep/nextflow/workflows/run_vep.nf"

workflow {
	meta = [:]
        meta.one_to_many = 0
        meta.index_type = "tabix"

       ch_vep = Channel.of( [meta, file("path.to.vcf.file"), file("path.to.vcf.file.inedx"), file("path.to.vep_config") ])
}
```

nextflow-vep should also be tested in CLI mode to see if one-to-one, one-to-many, and many-to-one scenario works.


dbolser:
Has this been tested?

nuno-agostinho (approved):
Hey Nakib, this PR seems to be working as intended!

All scenarios work when using the main workflow and also using the sub-workflow works as expected. Thanks!

nuno-agostinho:
Merged to `main` and `release/112`.

PR #1498, opened 2023-09-14
Fix custom annotation of breakends and non-supported SVs (exact mode)

nuno-agostinho:
ENSVAR-5610, fixes #1397

### Changelog

- Custom exact matching improvements:
  - Avoid matching breakpoints with SNPs when using custom annotation in exact mode
  - Only perform this check when using custom annotation in VCF format (other formats do not contain variant type information in a standardised way)
- Bug fix: avoid skipping all subsequent SVs after skipping one
- Improvements to warning messages thrown when skipping variants:
  - Warn when skipping any non-supported SV type
  - Trim input whitespaces to improve readability of messages
  - Bug fix: avoid calling `skipped_variant_msg()` with ensembl-io's parser

### Documentation

- Updates to public documentation: https://github.com/Ensembl/public-plugins/pull/722

### Testing

Run VEP with custom annotation in exact mode containing breakends and `NON_REF` SVs, such as:

```
1   2642609        .                               G  A  .  .  .  .  .
1 2642609   nonREF  G  <NON_REF>  892  PASS  AN=6  GT  0/1
chr11   99819752        nonREF       C       <NON_REF>        892     PASS    AN=6       GT    0/1
1       713044  DUP1     C       <CN=2>  .       .       END=755966
1       713044  DUP2     C       <CN2>  .       .       END=755966
1       713044  DUP2     C       <INS:FHJDSJH>  .       .       END=755966
chr11   99819752        MantaBND:2716:0:1:0:0:0:0       C       CAACTGAG]chr11:99820576]        892     PASS    SVTYPE=BND;MATEID=MantaBND:2716:0:1:0:0:0:1;SVINSLEN=7;SVINSSEQ=AACTGAG;BND_DEPTH=266;MATE_BND_DEPTH=24;AC=3;AN=6       GT    0/1
chr11   99820576        MantaBND:2716:0:1:0:0:0:1       C       CCTCAGTT]chr11:99819752]        892     PASS    SVTYPE=BND;MATEID=MantaBND:2716:0:1:0:0:0:0;SVINSLEN=7;SVINSSEQ=CTCAGTT;BND_DEPTH=24;MATE_BND_DEPTH=266;AC=3;AN=6       GT    0/1
```

Run VEP using a command such as:
```
gnomAD=input/gnomad.genomes.v3.1.2.sites.stripped.vcf.gz
vep --i $vcf --vcf --database --custom $gnomAD,gnomAD,vcf,exact,0,AF,HN
```

[...]

PR #1513, opened 2023-10-09
Fix unit tests in main branch

nuno-agostinho:
This change was introduced in https://github.com/Ensembl/ensembl-vep/pull/1498, but was lost in-between merge conflicts. Everything else seems fine when comparing with `postreleasefix/111` changes.


nakib103 (approved):
LGTM!~

PR #1517, opened 2023-10-12
Release/110 patch

dbolser:
I know this is pedantic, but the warnings caused by empty lines in the config file are really ugly!

I like a nice, easy to read config with plenty of space, so I get tons of warnigns!

Many thanks.


dbolser:
Gah, pr'd the wrong commits...

PR #1523, opened 2023-10-15
Update INSTALL.pl

dbolser:
Making CLI options mirror variables... for some reason.


dbolser:
Not sure what I did in the end, but this PR looks correct at least...

likhitha-surapaneni:
Merged to `main` and `release/111`

PR #1538, opened 2023-10-23
Fix Docker build badge

nuno-agostinho:


likhitha-surapaneni (approved):
Thanks @nuno-agostinho

PR #1548, opened 2023-11-03
Restrain using latest DBD::mysql version

nakib103:
DBD::mysql have released new version (5.001) on Oct 4, 2023. This version removes support for MySQL 5 -
https://metacpan.org/dist/DBD-mysql/changes

Hence we need to use lower version than that.

Ensembl repo already have similar PR merged - https://github.com/Ensembl/ensembl/pull/666


likhitha-surapaneni (approved):
LGTM

PR #1561, opened 2023-11-21
Parse CNV:TR alleles from VCF

nuno-agostinho:


modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
```diff
@@ -433,6 +433,70 @@ sub create_StructuralVariationFeatures {
...
+      my @RN = $info->{RN} ?
+        split(/,/, $info->{RN}) :
+        # if RN is missing, assume 1 repeat for each CNV:TR allele
+        (1) x ( () = $type =~ /CNV:TR/g );
+      my @RUS = split /,/,  $info->{RUS};
+      my @RUC = split /,/, ($info->{RUC} || ''); # undefined if no RUC
+      my @RB  = split /,/, ($info->{RB}  || ''); # undefined if no RB
+      my $is_RUC_defined = scalar @RUC;
+
+      my $max_sv_size = $self->param('max_sv_size') || 5000;
```
nakib103, line 456: As we are using VF to calculated consequence for this variant and `max_sv_size` can be very large depending on user preference - should we do a benchmark and put some statement in the doc if it reduce speed significantly?

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
```diff
@@ -433,6 +433,70 @@ sub create_StructuralVariationFeatures {
...
+          last if $is_oversized;
+          $repeat .= $seq x $num;
+        }
+
+        last if $is_oversized;
+        # prepend reference allele
+        push @repeats, $ref . $repeat;
+      }
+      # avoid storing alternative allele for tandem repeats with large repeats
+      $alt = $ref.'/'.join('/', @repeats) unless $is_oversized;
```
nakib103, line 480: the default value of `max_sv_size` is 10 million. In case there is a lots of repeat the allele string would be huge. Should we do something about it in the output file? Like keep the same parameters in the output VCF and same alt as input VCF. The other format would be a problem.

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
nuno-agostinho: I think we can ignore benchmark for now. This is currently just a spec and we will need to wait until real-world examples are available to properly test if this implementation needs refinement.

modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
```diff
@@ -433,6 +433,70 @@ sub create_StructuralVariationFeatures {
...
+
+      if (!$is_oversized) {
+        # convert tandem repeats to VariationFeature in order to properly
+        # calculate consequences based on alternative allele sequence
+        my $vf = Bio::EnsEMBL::Variation::VariationFeature->new_fast({
+          start          => $start,
+          end            => $start + length($ref) - 1,
```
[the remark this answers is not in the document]
nuno-agostinho: Fixed the reference to get sequence based on start and end (`SVLEN`) of the variant in [`6fcd5e5` and `d99dc66`](https://github.com/Ensembl/ensembl-vep/pull/1561/files/6fcd5e5e3306662a58a32a52469db39497ec56ce..d99dc66e053d53a85cdada36090a55c9a8d6ac5a).

modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
nuno-agostinho: Hey @nakib103, after thinking this through, I think this is more a VEP-wide issue, than specifically for this case (after all, the user can always input large tandem repeats without using `<CNV:TR>`).

The easiest way to represent tandem repeats would be using the notation in the PR description, e.g., displaying `CAGCAGCAGATAT` as `(CAG)3(AT)2`. However, this would require us to test if VEP, `filter_vep` and other downstream tools (`bcftools`?) can handle that output. I can open a ticket to look into this later, what do you think?

nakib103: @nuno-agostinho, yep, makes sense.

nuno-agostinho: Opened ticket ENSVAR-6142. Thanks!

modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
```diff
@@ -433,6 +487,51 @@ sub create_StructuralVariationFeatures {
...
+  elsif ($so_term =~ /tandem/ ) {
+    # validate SVLEN
+    if (defined $info->{SVLEN}) {
+      my @svlen = uniq(split(/,/, $info->{SVLEN}));
+      # warn if there are different references per alternative allele
+      $self->warning_msg(
+        "found tandem repeats with different references per alternative allele: " .
+        "SVLEN=" . $info->{SVLEN} . "; only using reference with largest size"
+      ) if scalar @svlen > 1;
+      $end = $start + max(@svlen) - 1;
```
nakib103, line 500: Hi, @nuno-agostinho ,
everything looks fine. Just here we should not put a minus 1.
```suggestion
      $end = $start + max(@svlen);
```

From[ VCF specification 4.4](https://samtools.github.io/hts-specs/VCFv4.4.pdf) page 14 -
```
symbolic structural variant alleles:, POS + SVLEN.
```

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/Parser/VCF.pm
nakib103: As pointed out by nuno, start is actually `POS + 1`.

(https://github.com/Ensembl/ensembl-io/blob/b1a0d57104ed2532e958194c38159a6092ba8b7b/modules/Bio/EnsEMBL/IO/Parser/BaseVCF4.pm#L262)

So the current logic is correct.

nakib103:
LGTM!

nakib103 (approved):
LGTM!

nakib103:
merged with `release/112` and `main`

PR #1575, opened 2023-12-12
Avert cross species contamination in VEP cache dump

nakib103:
[ENSVAR-6087](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6087)

#### Problem
In VEP cache we generally have `pubmed` and `var_synonyms` data from the database. For past couple of release we are getting erratically missing data from them. Note that these two are the only field that are queried from the database and dumped into files and read back again from those dump file when creating the cache.

#### Cause
After adding some debug data (the contents of the dump files read in each job), the cause behind these missing data is found out. We generally keep the data loaded from the dump files in-memory for a species between jobs. But it seems that somehow these objects is persisting jobs between different species and hence one species is getting data from another species.

### Solution
Simply do not keep in-memory objects. 

#### Test
Tested with fix and compared the result between different species. The result can be viewed in above JIRA ticket.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/Dumper/Variation.pm
+++ b/modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/Dumper/Variation.pm
@@ -429,68 +429,58 @@ sub pubmed {
   my $self = shift;
   my $as = shift;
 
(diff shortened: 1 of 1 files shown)


jamie-m-a (approved):
Agree with solution -  removed option to use an existing object and force creation of new object each time these subs called.

PR #1583, opened 2024-01-12
Update docker.yml

nuno-agostinho:
- Relax regex used to capture VEP version
- Add example of Docker image tag


likhitha-surapaneni (approved):
LGTM, thanks @nuno-agostinho

PR #1611, opened 2024-02-13
Fix unit test regarding HGVS dup lookup update

nakib103:
change related to https://github.com/Ensembl/ensembl-variation/pull/1042


nuno-agostinho (approved):
Thanks @nakib103, looks good to me!

PR #1673, opened 2024-05-14
Allow filtering in nextflow-vep

nakib103:
Added a param `--filters` which takes a string as an argument which can be comma separated list of filter condition. These condition are similar to what we can give in `filter_vep`.

As nextflow-vep can handle only VCF files the `filter_vep` arguments used here is also tailored for VCF file.

### Test Example

```
nextflow -C ${repositories}/ensembl-vep/nextflow/nextflow.config run ${repositories}/ensembl-vep/nextflow/workflows/run_vep.nf \
--vcf $input_file \
--vep_config $vep_config \
--outdir $output_dir \
--output_prefix $output_prefix \
--filters "amino_acids not match X[A-Za-z*]?\/,Feature is ENST00000377918"
```

example variant to check this filter
```
13      57725290        rs1160403026    AGTG    A
```

### DIFF
--- a/nextflow/nf_modules/run_vep.nf
+++ b/nextflow/nf_modules/run_vep.nf
@@ -30,6 +30,7 @@ process runVEP {
   script:
   index_type = meta.index_type
   out_vcf = "vep" + "-" + file(original_vcf).getSimpleName() + "-" + vep_config.getSimpleName() + "-" + vcf
+  tabix_arg = index_type == 'tbi' ? '' : '-C'
   
   if( !vcf.exists() ) {
     exit 1, "VCF file is not generated: ${vcf}"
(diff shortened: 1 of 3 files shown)


nextflow/nf_modules/run_vep.nf
```diff
@@ -38,6 +37,25 @@ process runVEP {
...
+    vep -i ${vcf} -o STDOUT --vcf --format vcf --config ${vep_config} | filter_vep -o filtered.vcf --only_matched --format vcf ${filter_arg}  
+    
+    bgzip filtered.vcf
+    mv filtered.vcf.gz ${out_vcf}
+    
+    if [[ "${index_type}" == "tbi" ]]; then
+      tabix -p vcf ${out_vcf}
+    else
+      tabix -C -p vcf ${out_vcf}
+    fi
```
nuno-agostinho, line 56: Let's simplify this Tabix statement.

How about we define the Tabix argument after defining `out_vcf`:
```nextflow
tabix_args = index_type == 'tbi' ? '' : '-C'
```

And replace these (and the next block's) if/else lines with:
```nextflow
tabix ${tabix_args} -p vcf ${out_vcf}
```

nextflow/workflows/run_vep.nf
nuno-agostinho, line 1: Add parameter usage to the help info below? Something like:

```
  --filters STRING        Comma-separated list of filter conditions to pass to
                          filter_vep, such as "amino_acids not match
                          X[A-Za-z*]?\/,Feature is ENST00000377918").
                          Default: filter_vep is not run
```

nuno-agostinho (changes requested):
Hi @nakib103, thanks for the changes!

I left some suggestions.

nextflow/nf_modules/run_vep.nf
nakib103: thanks @nuno-agostinho , good suggestion. updated!

nextflow/workflows/run_vep.nf
nakib103: Done 👍

nuno-agostinho (approved):
Working as expected, thanks @nakib103!

Issue #1646, opened 2024-03-27
Custom VCF w/overlap joins multiple FILTER values for a record with same delimiter ("&") as other records


davmlaw:
## Describe the issue

When ```--custom``` is used with overlap, it can return multiple records. INFO values from the multiple records are joined with "&"

Multiple FILTER values for the *SAME VCF record* are also converted to using the same delimiter ("&") which means you lose track of what record it came from. 

For instance if you annotate the test VCF provided against an annotation file:

```
chr1	13203398	gnomAD-SV_v3_DEL_chr1_cf5f2df8	N	<DEL:ME:LINE1>	348	HIGH_NCR;UNRESOLVED
chr1	13203534	gnomAD-SV_v3_DEL_chr1_52ee315d	N	<DEL:ME:LINE1>	175	UNRESOLVED
```

It outputs the equivalent of:

```
gnomAD_SV=gnomAD-SV_v3_DEL_chr1_cf5f2df8&gnomAD-SV_v3_DEL_chr1_52ee315d
gnomAD_SV_FILTER=HIGH_NCR&UNRESOLVED&UNRESOLVED
```

There should be 2 values for every field, but in gnomAD_SV_FILTER there are 3

This means that you can't match FILTER records to other records as the array sizes are different.

### System
- VEP version: 110

### Full VEP command line
```
vep -i input_sv.vcf -o vep_out.vcf --cache --dir /data/annotation/VEP/vep_cache --fasta /data/annotation/fasta/Homo_sapiens.GRCh38.dna.toplevel.fa.gz --assembly GRCh38 --offline --use_given_ref --vcf --force_overwrite --flag_pick --exclude_predicted --no_stats --check_existing --no_escape --sift b --hgvs --symbol --numbers --domains --canonical --protein --biotype --transcript_version --af --pubmed --variant_class --distance 5000 --refseq --custom file=gnomad_sv_grch38_example.vcf.gz,short_name=gnomAD_SV,format=vcf,type=overlap,overlap_cutoff=80,same_type=1,fields=AC%AF%FILTER
```

Attached ZIP containing input VCF, output VCF and 2 record custom annotation VCF w/multiple filters for 1 recrd

[custom_vcf_join.zip](https://github.com/Ensembl/ensembl-vep/files/14766251/custom_vcf_join.zip)

nakib103:
Hello @davmlaw,

[...]

PR #1689, opened 2024-06-03
Update delimiter for FILTER when used for custom option

nakib103:
[ENSVAR-6276](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6276)
https://github.com/Ensembl/ensembl-vep/issues/1646

When we use `FILTER` field from a VCF file in `--custom` argument, the delimiter for FILTER, which is `;`, is replaced by `,` ([see](https://github.com/Ensembl/ensembl-vep/blob/31a3581b84495b617b2f3980da6c6313ca6d238f/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm#L219)). The reason is `;` have specific meaning in certain file type - for example, in VCF where this `FILTER` is going to put under `INFO`,`;` is used as delimiter to split up different fields in it.

But we also replace any `,` with `&` if found within any specific value under `CSQ` INFO field ([see](https://github.com/Ensembl/ensembl-vep/blob/31a3581b84495b617b2f3980da6c6313ca6d238f/modules/Bio/EnsEMBL/VEP/OutputFactory/VCF.pm#L373)). That is because in `CSQ` field `,` is used as delimiter to separate different records. 

And thus a `;` can become a `&` in VCF CSQ field. Which causes the confusion mentioned in the issue above.

To avoid error replace `;` with `%3B` which we are already doing for VCF file ([see](https://github.com/Ensembl/ensembl-vep/blob/31a3581b84495b617b2f3980da6c6313ca6d238f/modules/Bio/EnsEMBL/VEP/OutputFactory/VCF.pm#L374)).

**Cons:** 
Other output format (tab, vep text) will now have `%3B` instead of `,` as delimiter.
To avoid that, according to @nuno-agostinho's suggestion utilized  [convert_arrayref](https://github.com/Ensembl/ensembl-vep/blob/f167c3a4656b974f41f932a51217d347909d2c81/modules/Bio/EnsEMBL/VEP/Utils.pm#L185) function to keep the `;` and let the OutputFactory method handle the delimiter substitution. Default VEP format needed to replace `;` by `%3B` too.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/VCF.pm
@@ -216,7 +216,7 @@ sub _create_records {
     }
     ## extract pass/fail info from filter column
     $fields_data->{FILTER} .= $parser->get_raw_filter_results();
-    $fields_data->{FILTER} =~ s/\;/\,/g;
+    $fields_data->{FILTER} =~ s/\;/\%3B/g; # a semi-colon have special meaning in VCF INFO column
   }
 
 
--- a/modules/Bio/EnsEMBL/VEP/OutputFactory/VEP_output.pm
+++ b/modules/Bio/EnsEMBL/VEP/OutputFactory/VEP_output.pm
@@ -113,7 +113,11 @@ sub output_hash_to_line {
 
   push @line, (
     join(';',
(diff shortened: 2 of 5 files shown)


nuno-agostinho (approved):
Thanks @nakib103, everything is working as expected!

PR #1706, opened 2024-06-24
Skip adding EMER regulatory feature in VEP cache

nakib103:
[ENSVAR-6367](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6367)


nuno-agostinho (approved):
Thanks @nakib103, looks good to me!

PR #1707, opened 2024-06-26
Input full path to cache directory

nuno-agostinho:
Add new `--full_cache_dir` option to set cache directory directly and check if given directory exists

@nakib103, should we improve the logic to check if the directory is a VEP cache dir? How would we do this?

## Testing

Run VEP with:
- `--full_cache_dir` with a non-existing directory: should raise error
- `--full_cache_dir` with a VEP cache dir: should work as expected
- `--cache_dir` and `--full_cache_dir` with a VEP cache dir: should ignore `--cache_dir`
- `--cache_dir` as usual: should work as expected

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/CacheDir.pm
+++ b/modules/Bio/EnsEMBL/VEP/CacheDir.pm
@@ -104,7 +104,9 @@ sub new {
 
   # need to either specify full path (dir) e.g. ~/.vep/homo_sapiens/78_GRCh38/
   # or root path (root_dir) e.g. ~/.vep/ where species and version are drawn from config and assembly from config, DB or scanning dir
-  throw("ERROR: No root_dir or dir specified") unless $hashref && ($hashref->{root_dir} or $hashref->{dir});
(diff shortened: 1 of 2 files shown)


nakib103 (commented):
Hi @nuno-agostinho, 

Everything is working as expected, thanks!

As for the checking if the `--full_cache_dir` is actually cache dir - we do not check for that with `--dir_cache` either. It would be nice but not mandatory I would say. User currently get warning -
```
variant .. not found in annotation sources or synonyms
```
But a clear warning of it happened because of faulty cache directory would be better.

We can add checking of presence of `info.txt` file [here](https://github.com/Ensembl/ensembl-vep/blob/31a3581b84495b617b2f3980da6c6313ca6d238f/modules/Bio/EnsEMBL/VEP/CacheDir.pm#L364) and further checking if species and assembly matches if present in config.

nakib103 (approved):
Checking if the cache dir is a proper is not mandatory. Something that can be added later on. LOTM!~

Issue #1705, opened 2024-06-20
Synonyms file does not work in the offline mode


XinmengLiao:
## Describe the issue
Hi. I am currently using VEP v111.0 to annotate my vcf files. When I use --offline mode and provided the synonyms file, errors shows up and some of the chromosomes do not overlap any features. 

### System
- VEP version: [111] 
- VEP Cache version: [111]
- Perl version: [5.26]

### Full VEP command line
```
vep --cache --dir_cache $VEP_CACHE \
--offline \
--fork 128 \
--format vcf \
--dir_plugins $VEP_plugins111/ \
-i Sample1_PASS.vcf.gz \
-o Sample1_vep_annotated.vcf.gz \
--force_overwrite \
--compress_output bgzip \
--assembly GRCh38 \
--symbol --vcf --check_existing --variant_class \
--sift b --polyphen b \
--synonyms $VEP_CACHE/homo_sapiens/111_GRCh38/chr_synonyms.txt \
--hgvs \
--fasta Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz \
--canonical \
--af --af_gnomade --af_gnomadg --max_af \
--custom clinvar_20240611_PLPC.vcf.gz,ClinVar,vcf,exact,0,ID,CLNSIG,CLNDN,CLNHGVS,CLNSIGINCL,CLNVC,GENEINFO,CLNDISDB,CLNSIGCONF,CLNREVSTAT,CLNDNINCL,CLNREVSTAT 
```

### Full error message
> WARNING: Line 249593 skipped (chr10_GL383545v1_alt 3485 . C CT 9.03 PASS AC=...): Chromosome chr10_GL383545v1_alt not found in annotation sources or synonyms; chromosome chr10_GL383545v1_alt does not overlap any features
WARNING: Line 249594 skipped (chr10_GL383545v1_alt 5295 . G GA 11.56 PASS AC...): Chromosome chr10_GL383545v1_alt not found in annotation sources or synonyms; chromosome chr10_GL383545v1_alt does not overlap any features
WARNING: Line 249710 skipped (chr10_KI270824v1_alt 73879 . T C 47.19 PASS AC...): Chromosome chr10_KI270824v1_alt not found in annotation sources or synonyms; chromosome chr10_KI270824v1_alt does not overlap any features
[...]

PR #1718, opened 2024-07-09
Fix missing non reference seq region in VEP cache

nakib103:
[ENSVAR-6377](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6377)
https://github.com/Ensembl/ensembl-vep/issues/1705

In e110, `assembly_exceptions` table was truncated which contained the sequence region for patches, haploptypes etc. due to changed mechanism of how those regions are showed in Ensembl genome browser.

The VEP cache pipeline depended on the `assembly_exceptions` table to grab those region. This PR aims to fix these issue by updating the current correct way to add non reference sequence regions.

### Test
pipeline url - http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-var-prod-4&port=4694&dbname=snhossain_dump_vep_human_113_test&passwd=xxxxx

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/CreateDumpJobs.pm
+++ b/modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/CreateDumpJobs.pm
@@ -166,8 +166,7 @@ sub get_chr_jobs {
 
   # get slices
   my $sa = $dba->get_SliceAdaptor;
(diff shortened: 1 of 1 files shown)


nuno-agostinho (approved):
Thanks for the PR, @nakib103! Everything seems fine and the resulting VEP cache includes the sequence regions missing from previous releases.

PR #1720, opened 2024-07-12
VEP 88 Docker: Create docker.yml worfklow

nuno-agostinho:
Add workflow to automatically create Docker


likhitha-surapaneni (approved):
LGTM, thanks @nuno-agostinho

PR #1727, opened 2024-07-25
DumpVEP: fix QC step for using non-reference sequence

nakib103:
Related PR: https://github.com/Ensembl/ensembl-vep/pull/1718

In above PR we updated the pipeline to consider non-reference sequence. We also need to update the QC step so it does not fail.

A pipeline w/t fix:
http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-var-prod-4&port=4694&dbname=snhossain_dump_vep_human_113&passwd=xxxxx


likhitha-surapaneni (approved):
Thanks @nakib103, qc is successful in the updated pipeline

PR #1733, opened 2024-08-07
Add pick_count option - number of results to return by pick

nakib103:
https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6422


nakib103:
Unnecessary; the goal can be achieved by `--transcript_filter` option.

PR #1735, opened 2024-08-08
Update slurm profile and resource limits

veidenberg:
This PR updates the slurm profile based on testing on our slurm cluster:
* Enable singularity on slurm
* Increase memory limit for VEP (for running without caches)

### DIFF
--- a/nextflow/nextflow.config
+++ b/nextflow/nextflow.config
@@ -39,6 +39,8 @@ profiles {
   }
   slurm {
     process.executor = 'slurm'
+    singularity.enabled    = true
+    singularity.autoMounts = true
   }
 
   conda {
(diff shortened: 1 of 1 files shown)


veidenberg:
@nuno-agostinho @nakib103 these updates enabled us to run the pipeline on our cluster. Please take a look whether it's worth merging (for some reason I couldn't add reviewers).

nuno-agostinho:
Hi @veidenberg!

@nakib103 and I think that the `slurm` profile should not include the Singularity configuration, as a user may want to use Docker instead when using that profile.

Instead, we propose to create a new profile with the website-specific configuration, like so:

```
ensembl {
  process.executor = 'slurm'
  singularity.enabled    = true
  singularity.autoMounts = true

  withName: runVEP {
    memory = 32.GB
  }
}
```

What do you think?

veidenberg:
Makes sense. Config updated and tested.

nuno-agostinho (approved):
Thanks for the changes, @veidenberg. Looks good to me!

PR #1737, opened 2024-08-08
Tiny bugfix

veidenberg:
Small typo that I noticed affecting our pipeline output. The quotes-inside-quotes default value for output filename prefix is not an empty value, resulting in all output files to be named `_VEP.vcf.gz` (instead of `<input-name>_VEP.vcf.gz`). This tiny edit would fix it.


veidenberg:
@nuno-agostinho @nakib103 I assumed it's a bug. Feel free to merge.

nakib103 (approved):
LGTM!~

PR #1738, opened 2024-08-09
Check extension also when deciding runVEPonVCF process

nakib103:
In nextflow-vep pipeline we decide on extension of the file and the header content to decide if we go to `runVEPonVCF` process. Added back the extension checking.


nuno-agostinho (approved):
Looks good to me, thanks @nakib103!

PR #1744, opened 2024-08-20
Capture uploaded allele correctly for VCF input

likhitha-surapaneni:
Ticket: [ENSVAR-5858](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5858)

- [x] Fix Unit tests
- [x] Review header and output columns

To test:

- [x] With `--allele_number` for multi-allelic examples

Testing:
```
1   961320  .   G   GCAGC,GCA
1   961323  .   G   GG,-
```


modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -976,7 +981,7 @@ sub minimise_alleles {
         $new_vf->{end}                    = $end;
         $new_vf->{seq_region_start}       = $start;
         $new_vf->{seq_region_end}         = $end;
-        $new_vf->{original_allele_string} = $vf->{allele_string};
+        $new_vf->{original_allele_string} = $vf->{original_allele_string} || $vf->{allele_string};
```
likhitha-surapaneni, line 984: Can we assign `$vf->{original_allele_string}` only if VCF?

nakib103: The way you are doing should be okay. For other parsers `$vf->{original_allele_string}` is not defined so it will take the value from `$vf->{allele_string}`. 

To only assign only if VCF we can check the parser type is `Bio::EnsEMBL::VEP::Parser::VCF` but seems redundant.

modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -950,10 +950,15 @@ sub minimise_alleles {
```
nakib103, line 950: I think there is an error in L859 where we determine the `$is_indel`. 

Say for a VCF input `1 100 . A AG` the `$ref_allele_string` and `$alt_allele_string` will be `-\G` and thus length is same. Maybe we should have a check if `$ref_allele_string` or `$ref_allele_string` have `-`.

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -950,10 +950,15 @@ sub minimise_alleles {
...
     # skip VFs with more than one alt
     # they get taken care of later by split_variants/rejoin_variants
-    if(!$vf->{allele_string} || $vf->{allele_string} =~ /.+\/.+\/.+/ || $vf->{allele_string} !~ /.+\/.+/) {
+    if(!$vf->{allele_string} || $vf->{allele_string} !~ /.+\/.+/) {
       push @return, $vf;
     }
 
+    elsif($vf->{allele_string} =~ /.+\/.+\/.+/)
+    {
+      $vf->{minimised} = 1;
```
nakib103, line 959:
```suggestion
      push @return, $vf;
      $vf->{minimised} = 1;
```

Also need to return the $vf otherwise you get undef and empty results.

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/Parser.pm
likhitha-surapaneni: Hi @nakib103 ,
We are checking if the allele_string contains `-` [here](https://github.com/Ensembl/ensembl-vep/blob/bf0d2cf9e9d12a6512a15805e712554760793036/modules/Bio/EnsEMBL/VEP/Parser.pm#L859), are you referring to an additional check?

nakib103: Ah, right. You are already checking it. I think it is the condition that needs to change -
The `or` just needs to be changed to `and` and the `=~` needs to change to `!~`.

likhitha-surapaneni: We also do not want to call minimise_alleles if it is already minimised such as the case in `-/G`, maybe the variable name can be changed to `is_non_minimised_indel` or do you think there is any case we might be missing?

nakib103: True, we do not need to pass the `-/G` through minimisation. Changing the variable name might make it clearer. I was evaluating the condition to check for indels.
thanks @likhitha-surapaneni

modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -851,13 +851,24 @@ sub post_process_vfs {
...
+      my $is_non_minimised_indel = 0;
+      # For VCF input, the allele_string is trimmed to remove anchor base, so we capture original allele
+      
+      my $original_allele_string = ($vf->{nontrimmed_allele_string} || $vf->{allele_string});
+      my @alleles = split(/\//, $original_allele_string);
+      my $ref_allele_string = shift @alleles;
+      my $alt_allele_count;
+
+      foreach my $alt(@alleles) {
+        if (length($ref_allele_string) != length($alt) or $original_allele_string =~ /^-/){
```
nakib103, line 866: what happens to `G/-/GGG` variant?
also `-/G`?

first should not go to minimise - should not be a problem
second one would not go to minimise - nothing would be done. so worthless.

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/OutputFactory.pm
```diff
@@ -2193,7 +2194,8 @@ sub rejoin_variants_in_InputBuffer {
   foreach my $vf(@{$buffer->buffer}) {
 
     # reset original one
-    if(defined($vf->{original_allele_string})) {
+    # check $vf->{first} to get first $vf in split_variants
+    if(defined($vf->{first})) {
```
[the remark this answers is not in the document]
likhitha-surapaneni: The way rejoin_required works is, 
it relies on a first variant_allele object which acts as an anchor for the variant and joins all the subsequent variant alleles to the anchor object. The proxy for first/anchor variant previously used to be $vf->{original_allele_string}. However with this[ line](https://github.com/Ensembl/ensembl-vep/blob/56d965a6dd5bf23205c17ae10639a40434f9d322/modules/Bio/EnsEMBL/VEP/Parser.pm#L860) , original_allele_string is defined for all variant alleles. We need another proxy variable for first variant, hence defined [$vf->{first}](https://github.com/Ensembl/ensembl-vep/blob/56d965a6dd5bf23205c17ae10639a40434f9d322/modules/Bio/EnsEMBL/VEP/InputBuffer.pm#L667)

modules/Bio/EnsEMBL/VEP/Parser.pm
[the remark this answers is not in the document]
likhitha-surapaneni: [$original_allele_string](https://github.com/Ensembl/ensembl-vep/blob/56d965a6dd5bf23205c17ae10639a40434f9d322/modules/Bio/EnsEMBL/VEP/Parser.pm#L860) contains both ref_allele and alt_alleles in the same string separated by "/"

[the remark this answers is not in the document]
nakib103: true, but with `^` you are only checking `-` allele for ref and missing deletions?

modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -950,7 +961,15 @@ sub minimise_alleles {
...
     # they get taken care of later by split_variants/rejoin_variants
-    if(!$vf->{allele_string} || $vf->{allele_string} =~ /.+\/.+\/.+/ || $vf->{allele_string} !~ /.+\/.+/) {
+    if(!$vf->{allele_string} || $vf->{allele_string} !~ /.+\/.+/) {
+      push @return, $vf;
+    }
+
+    elsif($vf->{allele_string} =~ /.+\/.+\/.+/)
+    {
+      # Updating a flag to minimise multi-allelic variants in split_variants/rejoin_variants
+      $vf->{minimised} = 1;
```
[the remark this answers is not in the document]
likhitha-surapaneni: Hi @nakib103 , can you please test this example with the latest commit. The allele is expected to be similar to when running `--minimal`

dglemos:
The `MINIMISED` is missing from the header, so in the VCF output even when the variant is minimised this info is not in the file. The only outputs with this info are the default and json.

dglemos:
When the input is in VCF format, example:
`1       961320  .       G       GCAGGCTCGGCC    .       .       .`

In the vep default output, the `Uploaded_variation` is represented with minimised alleles: 
`1_961321_-/CAGGCTCGGCC `
In this scenario matching the alleles is more difficult.

dglemos:
Yes, UPLOADED_ALLELE is enough to match the alleles. It should be returned by default though, if the variant is minimised it should return the original alleles too.

nakib103:
> Yes, UPLOADED_ALLELE is enough to match the alleles. It should be returned by default though, if the variant is minimised it should return the original alleles too.

I think it is generally advisable to not update the default output list so not to break existing pipelines.

modules/Bio/EnsEMBL/VEP/Parser.pm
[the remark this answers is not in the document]
likhitha-surapaneni: Hi @nakib103 , the output is still minimsed if you notice the 3rd column (correct me if I got it wrong). We have the allele "-" as the minimised representation for first alternative allele is GG/- and for the second one is G/-. The problem however is that there is no way to differentiate between the alternative alleles as both show "-" which is also a existing problem with the option `minimal` and needs to be addressed probably in a future ticket. Let me know if this makes sense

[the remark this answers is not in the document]
nakib103: I was actually looking at the `Uploaded_variation` column, it does not seem to be minimised as it does for bi-allelic variants (see - `1 961320 961324 GCAGG/GCA +`).

But the `Allele` column shows the alleles are minimised. It seems Uploaded_variation string has different logic, should we address this?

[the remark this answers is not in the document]
likhitha-surapaneni: Yes, I agree @nakib103. The uploaded variation was not expected to minimise alleles but it was minimising the alleles in some cases due to the way we populate the column and previously it was difficult to correct this as it is also the default option, hence the approach was taken to capture original allele with a different flag (`--uploaded_allele`). However, I agree it may also be worth investigating why uploaded variation is minimising the allele in some cases and not doing it in other cases.

likhitha-surapaneni:
> The `MINIMISED` is missing from the header, so in the VCF output even when the variant is minimised this info is not in the file. The only outputs with this info are the default and json.

Currently, the header is formed and written to the output file before looking at the input: https://github.com/Ensembl/ensembl-vep/blob/a6786e4357f442a81624f58d9e79f343909d717f/modules/Bio/EnsEMBL/VEP/Runner.pm#L204-L209

nakib103 (approved):
Looks good!~

dglemos (approved):
@likhitha-surapaneni, the changes look good.

PR #1748, opened 2024-08-28
Docker: Raise error if VEP version and release branch mismatch

nuno-agostinho:
When the branch name to build the Docker image starts with `release/` (release branch), raise an error if there is a mismatch between VEP version and the version in the release branch name.

## Testing

Check Docker image build completes depending on the branch argument:
- If `ARG BRANCH=main`, it should build successfully
- If `ARG BRANCH=release/113`, it should pass (depending on the VEP version in [modules/Bio/EnsEMBL/VEP/Constants.pm](https://github.com/Ensembl/ensembl-vep/blob/main/modules/Bio/EnsEMBL/VEP/Constants.pm))
- If `ARG BRANCH=release/113.4`, it should pass (subversions are ignored)
- If `ARG BRANCH=release/112`, it should fail

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,6 +32,16 @@ WORKDIR $OPT_SRC
 # Add ensembl-vep files from current context
 ADD . ensembl-vep
 
+# For release branches, raise an error if VEP version does not match the branch name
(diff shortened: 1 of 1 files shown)


likhitha-surapaneni (approved):
Works as expected, thanks @nuno-agostinho

PR #1749, opened 2024-08-29
Nextflow: add parameter to specify VEP version

nuno-agostinho:
## Changelog

- New `--vep_version` parameter allows to specify the VEP version of the image downloaded from Docker Hub
- Add information on downloading VEP cache
- Improve documentation

### DIFF
--- a/nextflow/README.md
+++ b/nextflow/README.md
@@ -55,10 +55,10 @@ The following config files are used and can be modified depending on user requir
 ```bash
 [Mandatory]
   --input FILE              Input file (if unsorted, use --sort to avoid errors in indexing the output file). Alternatively, can also be a directory containing input files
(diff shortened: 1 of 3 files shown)


nextflow/README.md
```diff
@@ -42,24 +45,28 @@ The following config files are used and can be modified depending on user requir
...
+  -profile <profiles-to-use> \
   --input <path-to-file> \
-  -profile <standard or lsf or slurm>
+  --vep_config <path-to-vep-config>
 ```
 
 #### Options
 
 ```bash
+[Mandatory]
```
nakib103, line 56: @nuno-agostinho , `vep_config` also seems mandatory params now. I don't think we can run nextflow-vep without it? In that case beside this section we should also update the Example section.

nuno-agostinho: Yes, thanks for noticing, fixed!

nakib103 (approved):
LGTM!~

PR #1751, opened 2024-09-04
Do not install Devel::Cover if perl version is 5.10

nakib103:


nuno-agostinho (changes requested):
Thanks @nakib103, seems to be working! I just have one suggestion:

travisci/install_dependencies.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+if [[ "$COVERALLS" = 'true' && ! "${TRAVIS_PERL_VERSION}" =~ '5.10' ]]
+then
+  awk '!/Devel::Cover/' ensembl/cpanfile > tmp_cpanfile && mv tmp_cpanfile ensembl/cpanfile
+  awk '!/Devel::Cover/' cpanfile > tmp_cpanfile && mv tmp_cpanfile cpanfile
```
nuno-agostinho, line 6: Would be less verbose with `sed`:
```suggestion
  sed -i '/Devel::Cover/d' ensembl/cpanfile
  sed -i '/Devel::Cover/d' cpanfile
```

PR #1753, opened 2024-09-10
nextflow-vep: output_prefix can be null when used from workflow

nakib103:
We check if output_prefix == "" otherwise keep the output file name same as input. But we did not consider when the nextflow-vep called from another pipeline output_prefix can be null. This PR attempts to fix that.

Also add --force_overwrite in filter_vep otherwise it gets in the way of -resume.


likhitha-surapaneni (approved):
LGTM, thanks @nakib103

PR #1758, opened 2024-09-23
Remove deprecated funcgen function

nakib103:
https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6436

### Test

example variants - 
```
Y 26671671 . A G
Y 17430762 . G T
```

Check database annotation as it changes the database annotation source -
```
vep -i input.txt --database --force --regulatory 
```

It will give no output no output for `--cell_type` parameter (`CELL_TYPE=` in default VEP format)


likhitha-surapaneni (approved):
Thanks @nakib103 , LGTM

PR #1762, opened 2024-10-02
Parse gene and transcript version from GFF/GTF file

nakib103:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2775

Transcript version is always `1` when GFF file is used. It seems it has been hard coded. This PR fixes this issue - 
- try to read version from GFF/GTF file
- if there is no version in the file reports no version even if `--transcript_version` is given.

### Test
Test with `rs699` and see if it matches transcript and gene version -
```
vep --id "1 230710048 rs699 A G" --force --gff /pathto/genes.gff3.bgz --fasta /pathto/Homo_sapiens.GRCh38.dna.toplevel.fa.gz --transcript_version --gene_version
```

Check the ticket for GFF and FASTA file location for testing.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/BaseGXF.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSource/File/BaseGXF.pm
@@ -478,12 +478,14 @@ sub _create_transcript {
     $tr_record->{strand}
   );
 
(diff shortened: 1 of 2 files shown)


likhitha-surapaneni (approved):
Thanks for this important fix @nakib103. Tested and works as expected

PR #1805, opened 2024-11-28
Allow https download and GITHUB_TOKEN in VEP Install

nakib103:
[ENSVAR-6431](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6431)

- Added option `--USE_HTTPS_PROTO` which would allow using https protocol to download cache and FASTA file instead of FTP. Useful in network where FTP port is blocked.
- Added option `--GITHUBTOKEN` or `--GITHUB_TOKEN` which would allow providing GITHUB token to query GitHub API using curl and allow increase in rate limit.

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -1637,6 +1637,7 @@ sub fasta() {
       }
     }
     elsif($USE_HTTPS_PROTO) {
+      print " - downloading $file\n" unless $QUIET;
       download_to_file("$FASTA_URL/$species/dna/$file", $ex);
     }
     else {
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -81,6 +81,7 @@ our (
   $NO_TEST,
   $NO_BIOPERL,
   $USE_HTTPS_PROTO,
+  $GITHUB_TOKEN,
   $ua,
 
   $CAN_USE_CURL,
@@ -180,7 +181,8 @@ GetOptions(
   'NO_HTSLIB|l',
   'NO_TEST',
   'NO_BIOPERL',
-  'USE_HTTPS_PROTO'
+  'USE_HTTPS_PROTO',
+  'GITHUBTOKEN|GITHUB_TOKEN=s',
 ) or die("ERROR: Failed to parse arguments");
 
 # Read configuration from environment variables starting with VEP_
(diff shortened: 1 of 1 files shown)


INSTALL.pl
```diff
@@ -1078,6 +1090,22 @@ ()
 # CACHE FILES
 #############
 
+sub convert_file_size {
+  # convert file size string to byte
+
+  my $size = shift;
+  my @units = ( 'K', 'M', 'G', 'T' );
+
+  $size =~ m/\s?(\d+\.?\d+)([T,G,M,K]?)/;
```
olaaustine, line 1099: Suggesting that this regex be changed to` m/\s?(\d+\.?\d+)([KMGT]?)/`

INSTALL.pl
```diff
@@ -1078,6 +1090,22 @@ ()
...
+
+  # in bytes already
+  return $1 unless defined $2;
+
+  # convert to bytes
+  my $scale = first {$units[$_] eq $2} 0..$#units;
+  return $scale ? $1 * (1024**($scale + 1)) : 0;
```
olaaustine, line 1106: `return defined $scale ? $1 * (1024**($scale + 1)) : 0; `

INSTALL.pl
```diff
@@ -1938,6 +2114,12 @@ sub usage {
                    a remote (e.g. FTP) address. The FASTA URL/directory must have
                    gzipped FASTA files under the following structure:
                    [species]/[dna]/
+--USE_HTTPS_PROTO  Download cache and FASTA file using HTTPs protocol instead of FTP.
+                   Useful for networks where FTP port is blocked by firewall.
+
+
+--GITHUBTOKEN      Set token to use for authentication when querying GitHub API. Authenticated
+                   user have increased rate-limit. PLEASE NOTE: use token with read-only access.
```
olaaustine, line 2122: Suggesting we change `PLEASE NOTE to NOTE `

olaaustine (commented):
Thanks @nakib103, just minor suggestions. Thank you!

INSTALL.pl
nakib103: done!

INSTALL.pl
nakib103: that's a good spot, fixed!

INSTALL.pl
nakib103: updated!

nakib103:
Hi @olaaustine , thanks for the good suggestions, have updated them all. Also, updated the usage text in public plugins doc - https://github.com/Ensembl/public-plugins/pull/844/files

olaaustine (approved):
LGTM

olaaustine:
Pushed to main and release/114

PR #1836, opened 2025-02-04
Update travis ubuntu version to focal

nakib103:
[ENSVAR-6409](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6409)

related ensembl-variation PR - https://github.com/Ensembl/ensembl-variation/pull/1160
related doc update - https://github.com/Ensembl/public-plugins/pull/865


.travis.yml
```diff
@@ -15,21 +15,20 @@ env:
...
       - apache2
       - libmysqlclient-dev
-      - libpng12-dev
+      - libpng-dev
       - libssl-dev
       - openssl
-      - mysql-client-5.6
-      - mysql-client-core-5.6
+      - mysql-client-5.7
+      - mysql-client-core-5.7
```
likhitha-surapaneni, line 31: Can we replace mysql dependencies with the following and try :
mariadb-server-10.6 
mariadb-client-10.6 
mariadb-client-core-10.6

nakib103: Thanks @likhitha-surapaneni!
Good point, I was looking into installing `5.7` version but for unittest purpose we can probably less stricter on matching version.

likhitha-surapaneni (approved):
LGTM, thank you for this change @nakib103

likhitha-surapaneni:
Hi @nakib103 , I get the following warning although everything is working as expected:

```
MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.
```

nakib103:
Hi @likhitha-surapaneni,

It is a false warning. `DBD::mysql` always set that option to `False` and you cannot change it. 
The warning comes from libmysqlclient version of ubuntu focal and we can ignore it. 

ref - https://github.com/perl5-dbi/DBD-mysql/issues/354

PR #1838, opened 2025-02-12
Remove redundant print in multi-file custom

nakib103:
There is a rogue print in custom annotation implementation of ingesting multiple files.


jamie-m-a (approved):
Nice catch - approved.

Issue #918, opened 2021-01-22
VEP doesn't copy headers from custom annotation files


mkohailan:
Hi,

I am using `--custom` annotations with VEP. The run goes well but the headers that define the annotations are written by VEP, which refers to the field of the annotation in the source file (i.e. it does NOT explain what that annotation is). Here is an example:

**Source annotation file header:**

##INFO=<ID=AF_AFR,Number=A,Type=Float,Description="Allele Frequency among African/African American genotypes, for each ALT allele, in the same order as listed">
##INFO=<ID=AF_AMR,Number=A,Type=Float,Description="Allele Frequency among Admixed American genotypes, for each ALT allele, in the same order as listed">
##INFO=<ID=AF_ASJ,Number=A,Type=Float,Description="Allele Frequency among Ashkenazi Jewish genotypes, for each ALT allele, in the same order as listed">

**VEP output file header:**

##INFO=<ID=gnomAD,Number=.,Type=String,Description="database/GnomAD/gnomad.genomes.r2.0.1.sites.noVEP.vcf.gz (overlap)">
##INFO=<ID=gnomAD_AF_AFR,Number=.,Type=String,Description="AF_AFR field from database/GnomAD/gnomad.genomes.r2.0.1.sites.noVEP.vcf.gz">
##INFO=<ID=gnomAD_AF_AMR,Number=.,Type=String,Description="AF_AMR field from database/GnomAD/gnomad.genomes.r2.0.1.sites.noVEP.vcf.gz">
##INFO=<ID=gnomAD_AF_ASJ,Number=.,Type=String,Description="AF_ASJ field from database/GnomAD/gnomad.genomes.r2.0.1.sites.noVEP.vcf.gz">

Any solution for this?

helensch:
Hi

Thank you for your query and providing the headers from the source annotation file and VEP output file.

The custom annotation adds a description using information from the custom annotation options.

https://www.ensembl.org/info/docs/tools/vep/script/vep_custom.html#custom_options

For your output the description includes the VCF field and source annotation filename.

This is similar to the output in the example of custom annotation for ClinVar:

https://www.ensembl.org/info/docs/tools/vep/script/vep_custom.html#custom_example

I will update the documentation to clarify the field descriptions that are used in the output of
custom annotations. 

I will discuss with the team about the inclusion of descriptions from custom annotation source VCF and get back to you.

Regards
Helen

mkohailan:
Great. Thanks!

[...]

PR #1853, opened 2025-03-20
Fix/propagate custom header

ainefairbrother:
## Description

This PR fixes the issue reported in #918 where the INFO header fields for custom annotation VCF files (supplied via the `--custom flag`) were not displaying the actual description from the custom file’s header. Instead, the output VCF showed a generic masked file path description. This PR addresses this such that the output VCF header now displays the correct INFO field descriptions as specified in the header of the file supplied to `--custom`.

## Changes to the PR

 **1.**  **AnnotationSource/File/VCF.pm**
- **Previously:** I'd added the \_extract_info_descriptions helper and called it from the constructor sub. 
- **Why this caused an issue?**: At that point the file hadn’t been set up with its real parser, so I was repeatedly hitting the fallback parser (return \$self->{parser} ||= Bio::EnsEMBL::IO::Parser::VCF4Tabix->open($self->file);). An additional test failed as it didn't expected the new field_descriptions key in the custom‐info object.
- **Now**: I've removed this logic from AnnotationSource/File/VCF.pm, reverting it back to it's state on main branch - this logic now acts later on in the OutputFactory VCF module. 

**2.** **VEP/OutputFactory.pm**
- **The changes here remain** - this logic chooses whether a real or generic description is propagated.

**3.** **VEP/OutputFactory/VCF.pm**
- **This is a new change:** OutputFactory is where VEP assembles its VCF output headers - I have added, at the top of the headers method, a loop that opens any vcf/vcf.gz/vcf.bgz and stores any field descriptions. 
- **Why this change?**:
    - This means that custom header extraction is now carried out as the output is being prepared, and after the correct parser has been assigned.

## Use case

When a VCF file is supplied to the `--custom` flag, the expected behaviour is for the INFO header lines in the output VCF to reflect the field descriptions provided in the custom file’s header. For example, if the custom file has an INFO field defined as:

```
##INFO=<ID=gnomAD_AF_ASJ,Number=A,Type=Float,Description="Allele Frequency among Ashkenazi Jewish genotypes, for each ALT allele, in the same order as listed">
```

then the output VCF header should use that description. Prior to this fix, the output header would instead show:

```
##INFO=<ID=gnomAD_AF_ASJ,Number=.,Type=String,Description="AF_ASJ field from [PATH]/gnomad.genomes.v4.1.sites.chr2.vcf.bgz">
```

This PR addresses that discrepancy by correctly parsing and propagating the real description.

## Benefits

- The output VCF now reflects the actual INFO field descriptions from the custom file - this will reduce confusion, as the VEP output file will directly reflect the source file.

## Possible Drawbacks

- Minor changes in header formatting may require additional testing with various custom file formats to ensure compatibility.

## Testing

- Travis tests passed locally
- I used the following test to confirm the new behaviour: 

```
cd /hps/software/users/ensembl/variation/fairbrot/ensembl-vep/

./vep \
    --input_file /hps/nobackup/flicek/ensembl/variation/fairbrot/tickets/6365_918/chr2_subset_100_NA12878.vcf.gz \
    --output_file "/hps/nobackup/flicek/ensembl/variation/fairbrot/tickets/6365_918/reproduce.vcf" \
    --format vcf \
    --vcf \
    --offline \
    --cache \
    --cache_version 113 \
    --dir_cache /nfs/production/flicek/ensembl/variation/data/VEP/tabixconverted \
    --assembly GRCh38 \
    --fasta /nfs/production/flicek/ensembl/variation/data/Homo_sapiens.GRCh38.dna.toplevel.fa.gz \
    --verbose \
    --force_overwrite \
    --custom file=/hps/nobackup/flicek/ensembl/variation/fairbrot/tickets/6365_918/gnomad_small_test.vcf.gz,short_name=gnomAD,fields=AC_afr%AF_afr%AN_afr,format=vcf,type=exact,coords=0

```
- The output file should contain the following INFO fields: 

```
##INFO=<ID=gnomAD_AC_afr,Number=.,Type=String,Description="Alternate allele count for samples in the African/African-American genetic ancestry group">
##INFO=<ID=gnomAD_AF_afr,Number=.,Type=String,Description="Alternate allele frequency in samples in the African/African-American genetic ancestry group">
##INFO=<ID=gnomAD_AN_afr,Number=.,Type=String,Description="Total number of alleles in samples in the African/African-American genetic ancestry group">
```


ainefairbrother:
TODO: I intend to test other custom formats to check that this fix has no off-target effects, though it is supposed to target VCFs only.

PR #1859, opened 2025-04-03
Remove directory for non-indexed cache

likhitha-surapaneni:
[ENSVAR-6629](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6629)

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/QCDump.pm
+++ b/modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/QCDump.pm
@@ -141,6 +141,7 @@ sub qc {
 
   # clean up
   rmtree($qc_dir);
+  rmtree($dump_dir.'/vep');
 
   return \@report_files;
 }


modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/QCDump.pm
```diff
@@ -141,6 +141,7 @@ sub qc {
 
   # clean up
   rmtree($qc_dir);
+  rmtree($dump_dir.'/vep');
```
nakib103, line 144: @likhitha-surapaneni,

It seems the `distribute_dumps` step uses the `vep` directory -
https://github.com/Ensembl/ensembl-vep/blob/2d6e9978e2221f34202a995417784a2e1d3fbf98/modules/Bio/EnsEMBL/VEP/Pipeline/DumpVEP/DistributeDumps.pm#L139

It copies unindexed cache file for species where indexed cache file is not present. I really do not where it is relevant.

Should we move this action further downstream? We can make the production wrapper to not copy for the `vep` directory here -
https://github.com/Ensembl/ensembl-production/blob/2916c1e8d16370f5b9f60b4b8032d2fb4c288c08/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/FileDumpVEP_conf.pm#L99

nakib103 (changes requested):

PR #1891, opened 2025-07-03
Patching test databases

jamie-m-a:
Updated for 115, few extra commits because of weird spacing


jamie-m-a:
> @jamie-m-a , the update to t/test-genome-DBs/homo_vepiens/variation is missing

Thanks Nakib - I didn't need to update, they were already there, see: https://github.com/Ensembl/ensembl-vep/blob/main/t/test-genome-DBs/homo_vepiens/variation/meta.txt

I did have the latest main, I pulled fresh for doing this PR, not sure what's happened. I think you can go ahead and merge if you're happy with it though

PR #1893, opened 2025-07-05
Custom annotation: prevent rogue FILTER output

nakib103:
`--custom` annotation outputs `FILTER` value from VCF file even if it is not mentioned in the [`fields`](https://www.ensembl.org/info/docs/tools/vep/script/vep_custom.html#custom_options).
This PR checks if `FILTER` is mentioned in the `fields` before adding it to the output.

**Changing of exisiting output**:
Although above behavior is undesirable but it is in existing output and can break pipelines. So I have added a internal-use argument `--custom_suppress_filter`, when used, would exercise the above mentioned correction.

This is used in the following Ensembl REST PR - 
https://github.com/Ensembl/ensembl-rest/pull/691

### Test
Any variant but needs to be annotated with VCF file and have `fields` in the custom args. Try running with and without `FILTER` in the `fields` and using `--custom_suppress_filter` argument.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Config.pm
+++ b/modules/Bio/EnsEMBL/VEP/Config.pm
@@ -261,7 +261,7 @@ our @VEP_PARAMS = (
   'ucsc_assembly=s',         # required for phyloP, phastCons, e.g. use hg19 for GRCh37, hg38 for GRCh38
   'ucsc_data_root=s',        # replace if you have the data locally, defaults to http://hgdownload.cse.ucsc.edu/goldenpath/
   'custom_multi_allelic',    # prevents filtering of custom annotation data when comma separated lists are assumed to be allele specific
-  'custom_suppress_filter',  # prevents adding from VCF FILTER column by default if not specified in 'fields' arguement, for internal use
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/VEP/Config.pm
```diff
@@ -261,6 +261,7 @@ our @VEP_PARAMS = (
   'ucsc_assembly=s',         # required for phyloP, phastCons, e.g. use hg19 for GRCh37, hg38 for GRCh38
   'ucsc_data_root=s',        # replace if you have the data locally, defaults to http://hgdownload.cse.ucsc.edu/goldenpath/
   'custom_multi_allelic',    # prevents filtering of custom annotation data when comma separated lists are assumed to be allele specific
+  'custom_suppress_filter',  # prevents adding from VCF FILTER column by default if not specified in 'fields' arguement, for internal use
```
dglemos, line 264: You could change the sentence to:
`prevents adding the VCF FILTER column unless it is explicitly requested in the 'fields' argument, for internal use`

dglemos (commented):
Tested with and without FILTER in the custom fields - it looks good.
I only added one comment about the option description.

dglemos (approved):

PR #1896, opened 2025-07-07
nextflow-vep: consider file name containing dot (.)

nakib103:
When we have input file name with dot nextflow-vep remove all the text after the first dot to create the filename. We only only want to remove the extension.

e.g. - test.vcf.gz --> test.vcf_VEP.vcf.gz

[this PR does not consider if the file is bgzipped or not complications]

Support how client discovers VEP output file in beta site -
https://github.com/Ensembl/ensembl-web-tools-api/blob/main/app/vep/vep_resources.py#L119-L124

### Test
Any variant with file name ending with `.vcf` and `.vcf.gz`.

### DIFF
--- a/nextflow/modules/merge_VCF.nf
+++ b/nextflow/modules/merge_VCF.nf
@@ -32,7 +32,7 @@ process mergeVCF {
   one_to_many = meta.one_to_many
   output_dir = meta.output_dir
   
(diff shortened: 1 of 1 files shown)


likhitha-surapaneni (approved):
LGTM, thanks @nakib103

PR #1908, opened 2025-07-31
Typo: avoid warning for un-initialized string

nakib103:
Fixed a typo where a hashref field is used before checking existence. Does not cause failure but to avoid getting warning in the output.

**Test**
run `perl t/AnnotationSource_File_VCF.t`


dglemos:
Merged into release/115 and main

PR #1928, opened 2025-10-01
Remove mention of deprecated option

jamie-m-a:
--af_esp hasn't existed since 106, so shouldn't be in this hash


ainefairbrother (approved):
LGTM.

PR #1936, opened 2025-10-14
Nextflow: Sort VCF with no header

likhitha-surapaneni:
[ENSVAR-6859](https://embl.atlassian.net/browse/ENSVAR-6859)


nextflow/modules/check_VCF.nf
```diff
@@ -69,10 +69,11 @@ process checkVCF {
   if( params.sort ) {
     isGzipped = vcf.extension == 'gz'
     cat_cmd   = isGzipped ? "zcat ${vcf}" : "cat ${vcf}"
-    sort_cmd += "(${cat_cmd} | head -1000 | grep '^#'; ${cat_cmd} | grep -v '^#' | sort -k1,1d -k2,2n) > tmp.vcf; "
+    sort_cmd += "(${cat_cmd} | head -1000 | grep '^#'; ${cat_cmd} | grep -v '^#' | sort -k1,1V -k2,2n) > tmp.vcf; "
     sort_cmd += isGzipped ? "bgzip -c tmp.vcf > ${vcf}" : "mv tmp.vcf ${vcf}"
   }
   """
+  [ -f *.gz ] || grep -q "^##fileformat" ${vcf} || sed -i '1i ##fileformat=VCFv4.2' ${vcf}
```
nakib103: - Do we need to check `[ -f *.gz ]` here? The goal is to check if there is header in the file.

- Also, there is another problem, the goal of separating header and headerless process was because we can work with `bcftools` with header-ed vcf file. But bcftools requires two header line to be present -
```
##fileformat...
#CHROM...
```

So only checking and adding `##fileformat` while the other header line missing will cause failure in sub-sequent steps.

- finally, doing a complete `grep` will slow down this process many-fold for large vcf file (such as for human dbSNP). So it is better to do grep only on header lines.

nakib103 (changes requested):
@likhitha-surapaneni , a significant problem here is that `checkVCFheader` is ran before `checkVCF`.

So adding the separation of with-header and without-header process happens before we can add header in `checkVCF` process.

What we can do instead is change `checkVCFheader` to a function called `isVCF` and separate all VCF into a flow where `checkVCF` process can add the needed headers.

PR #1938, opened 2025-10-15
Fix UT - update test cache+DB with new pf matrix

nakib103:
Update test database and cache with new protein function matrix according to https://github.com/Ensembl/ensembl-variation/pull/1176.

Some of the sift score changed **in cache** because (or at least my best guess) a part of the update where calculated by running the pipeline (which now uses updated sift version (6.2.1) than old times).


nakib103:
No need for this PR anymore - as the changes are made in ensembl-variation.

Issue #1968, opened 2026-01-29
-a should imply -n


FliesLikeABrick:
## Describe the issue
--auto (-a) can hang if someone is intentionally pinned to a specific version.  This can hang the install/build process or cause it to exit prematurely, depending on the context in which it is being run (if unattended or if stdin is closed so no input can be read and it exits).

Users can work around this by specifying -n, however they would need to know to do so which may not be the case after a build pipeline is live/productionized and this breaks at a later time

## Additional information
This can be particularly misleading if someone pins their build/install/deployment process to a version that happens to be latest; then at a later point in time a new release comes out; builds/installs/deployments will start failing

### System
- VEP version: 112
- VEP Cache version: None / N/A
- Perl version: N/A
- OS: N/A
- tabix installed ? N/A

### Full VEP command line
N/A this is an install-time issue

### Full error message
Do you wish to exit so you can get updates (y) or continue (n):

### Data files (if applicable)
N/A

PR #1969, opened 2026-01-29
Hotfix/#1968 auto should imply no update

FliesLikeABrick:
Skipping update checks when --auto is specified/configured.  Proposed fix for #1968


likhitha-surapaneni (approved):
Tested and looks good to me. Thanks @FliesLikeABrick

Issue #1967, opened 2026-01-20
Input files with whitespace in the filepath are word-split and not found


Jolleboll:
## Describe the issue
I run vep with an input file whose filepath contains whitespace, causing a "file does not exist" crash:
"gzip: \<word-split filepath\>: No such file or directory"
My input file is compressed, with a .gz file extension. I'm guessing there's Bash/shell code calling gzip for decompression before vep's Perl code executes, and that Bash code is perhaps not quoting properly, causing the word-splitting?

### System
- VEP version: 113
- VEP Cache version: 113
- Perl version: Whatever the docker image is bundled with.
- OS: openSUSE 15.6
- tabix installed Yes (but should not matter since I'm running through Docker/Singularity)

### Full VEP command line
```
vep="/home/user/Programs/VEP/image/vep.sif"
vepdir="/scratch/vep_cache"
sname="${bigvcf%%.*}" ; sname="${sname##*/}"
export outfmt=vcf ; export outext=vcf
bigvcf="/run/media/user/My Book/Andreea_vcf/allchrms.postVQSR.filtered.vcf.gz"

singularity exec "$vep" vep --dir "$vepdir"\
	--force_overwrite\
	-i "$bigvcf"\
	-o output/"$sname"."$outext".gz\
	--format vcf\
	--"$outfmt"\
	--compress_output gzip\
	--species homo_sapiens\
	--assembly GRCh38\
	--cache\
	--everything\
	--check_existing\
	--safe\
	--plugin CADD,/scratch/vep_cache/homo_sapiens/CADD_bigfiles/hg38/whole_genome_SNVs.tsv.gz,/scratch/vep_cache/homo_sapiens/CADD_bigfiles/hg38/gnomad.genomes.r4.0.indel_inclAnno.tsv.gz\
	--plugin ClinPred,/scratch/vep_cache/homo_sapiens/ClinPred/ClinPred_hg38_sorted_tabbed.tsv.gz\
	--offline\
	--fork 14
```

### Full error message
```
gzip: /run/media/user/My.gz: No such file or directory
gzip: Book/Andreea_vcf/allchrms.postVQSR.filtered.vcf.gz: No such file or directory
gzip: /run/media/user/My.gz: No such file or directory
gzip: Book/Andreea_vcf/allchrms.postVQSR.filtered.vcf.gz: No such file or directory
```

### Data files (if applicable)
Any gzipped vcf on a filepath with whitespace should do.

[...]

PR #1972, opened 2026-02-09
Handle white space in input path

nakib103:
https://github.com/Ensembl/ensembl-vep/issues/1967

[ENSVAR-6967](https://embl.atlassian.net/browse/ENSVAR-6967)


likhitha-surapaneni (approved):
works as expected, thanks @nakib103

soliton-:
Note that this does not fix the shell injection. It just changes the error case to filenames with an apostrophe in it.

Not involving a shell would look something like:
```perl
  open my $fh, "-|", "gzip", "-dc", "$file" or die("ERROR: $!");
```

See https://perldoc.perl.org/perlipc#Safe-Pipe-Opens

Issue #1971, opened 2026-02-04
Cache param in VCF header line metadata leaks infrastructure details


heuermh:
The VCF header line added by VEP leaks infrastructure details via the `cache=` parameter

```
##VEP="v115.2" API="v115" time="2026-01-22 18:43:15" cache="/fusion/s3/.../115_GRCh38/homo_sapiens/115_GRCh38" ...
```

Is there an easy way to either not include this parameter the VCF header line, or mask it?  We are using Ensembl VEP via nf-core workflows written in Nextflow.

ainefairbrother:
Hi @heuermh, 

Fix now merged, hope this helps. 

Many thanks, 

@ainefairbrother

PR #1974, opened 2026-02-13
Add flag to turn off cache path leaking in header

ainefairbrother:
See ENSVAR-6971 & issue #1971.

This PR fixes the cache path leaking in the VEP output file header line.  Masking behaviour for cache in header off by default, `--mask_header_cache_path` switches masking on. 

Before: 
```
##VEP="v115.0" API="v115" time="2026-02-13 15:08:56" cache="/nfs/production/flicek/ensembl/variation/data/VEP/tabixconverted/homo_sapiens/115_GRCh38" 1000genomes="phase3" ....
```

After: 
```
##VEP="v115.0" API="v115" time="2026-02-13 15:08:57" cache="[PATH]/115_GRCh38" 1000genomes="phase3" ....
```

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/BaseRunner.pm
+++ b/modules/Bio/EnsEMBL/VEP/BaseRunner.pm
@@ -417,7 +417,7 @@ sub get_output_header_info {
       $info->{version_data}->{$_} ||= $as_info->{$_} for grep {$_ ne 'custom_info'} keys %$as_info;
       if($as->can('dir')) {
         my $cache_dir = $as->dir;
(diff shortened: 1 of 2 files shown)


nakib103 (changes requested):
@ainefairbrother , 

the original path masking logic is implemented here -
https://github.com/Ensembl/ensembl-vep/blob/aac4870f9106c28ee5cadd9d5ee41e09369552d0/modules/Bio/EnsEMBL/VEP/Config.pm#L732

I think we should check why it is not working instead of adding the code logic in separate place?

nakib103 (approved):
LGTM !~

PR #1987, opened 2026-03-20
Nextflow caching fix

mluypaert:
This PR fixes nextflow caching for a number of steps, including the VEP execution, which were not being cached when running nextflow with `-resume` flag.


mluypaert:
@nakib103 This PR is now ready for review (after adding a few changes to enable post-processing result mapping in the VCF prepper pipeline)

nextflow/workflows/run_vep.nf
```diff
@@ -133,49 +131,51 @@ workflow NF_VEP {
...
-  vep_config = createInputChannels(params.vep_config, pattern="*.ini")
+  def input = createInputChannels(params.input, "*")
+  def vep_config = createInputChannels(params.vep_config, "*.ini")
+
+  def input_count = input.count()
+  def config_count = vep_config.count()
+
+  if (input_count != 1 && config_count != 1) {
+    exit 1, "Detected many-to-many scenario between VCF and VEP config files - currently not supported"
+  }
```
likhitha-surapaneni:
```suggestion
input_count
    .combine(config_count)
    .map { i, c ->
          if (i != 1 || c != 1) {
            error "Detected many-to-many scenario between VCF and VEP config files"
          }
    }
```
When testing the pipeline in standalone mode outside vcf_prepper, it is erroring out due to possible lazy loading of variables

likhitha-surapaneni (changes requested):
Hi @mluypaert,

Thank you for fixing caching. tested within vcf_prepper and in standalone mode and have one suggestion

nextflow/workflows/run_vep.nf
mluypaert: The logic in the above suggestion wasn't entirely correct (testing if either input or config count is not equal to one rather than both), but the changes introduced in d1c32715 should fix this issue.

likhitha-surapaneni (approved):
LGTM, thanks for getting this @mluypaert

Issue #1645, opened 2024-03-25
Feature Requests for custom VCF to be able to retrieve both ID and coords (to replace StructuralVariantOverlap)


davmlaw:
I am trying to use --custom instead of StructuralVariantOverlap plugin

However it's a bit hard to retrieve the percent overlap

If you use ```coords=1``` you can retrieve the coords (thus work out the overlap percentage in a post-processing step) however then you can't get back the VCF ID field which I need to be able to link to the gnomAD site.

There is a work around if you pass --custom twice with the same VCF, though this is inefficient.

Suggestion:

* Perhaps you could pass "ID" similarly to how you pass "FILTER", then you could get back both ID and coords
* You already calculate percent overlap - if there was a way to pass this back as annotation, this would make custom feature complete with StructuralVariantOverlap and you could get rid of that plugin

davmlaw:
Thanks! No hurry - whenever you have time

sbaross:
Is there any news on this? The [examples](https://www.ensembl.org/info/docs/tools/vep/script/vep_example.html#StructVar) page states that all features of the StructuralVariantOverlap plugin have been ported to custom annotations but I can't see anyway of printing the % overlap yet? It would definitely make life a lot easier than calculating it in post-processing

nakib103:
Hi, the feature is not implemented yet. We will see if we can add it in the next release, thank you.

sbaross:
While I've been calculating the overlaps myself I think there's a bug in how insertions are being reported. Insertions have the same start & end position regardless of length, but when using `coords=1` they're printed as `chr:start-[start+length]` which is inaccurate, and it seems like this inaccurate end position is then being used to calculate the overlaps meaning two separate insertions at different locations can be counted as overlapping

nakib103:
Hi @sbaross, SV insertion coord are calculated based on POS and SVLEN/END and is been used in overlap calculations.

PR #1989, opened 2026-03-24
[Custom annotation] New option coords=2 and show_overlaps

nakib103:
https://github.com/Ensembl/ensembl-vep/issues/1645

### coords=2
Added option of add both ID and coords when coords=2. The coords will appear the same way. ID will be shown with field name `{short_name}_ID`.
This does not affect GFF annotation as they are fundamentally different (create separate line in the output).

### show_overlaps
We can now add `show_overlaps=1` and the output will contain `{short_name}_OverlapBP` and `{short_name}_OverlapPC` containing the overlap bp and percentage. By default they are not shown.

#### Test
```
#input variant
1 230710048 . T <DEL> . . SVLEN=10

# custom file
1       230710048      variant1   A       <DEL>   .       .       SVLEN=100;TEST=hello
```

Try setting `coord=2` and `show_overlaps=1`


jamie-m-a:
Putting a hold on this one, not working as intended. coords=2 adds new overlap base pairs and percentage.
- percentage varies across transcripts, isn't using correct input.
- show_overlaps isn't adding separate overlaps like it used to.
- coords=1 is now contaminated with the new overlaps PC and BP values.

Issue #1990, opened 2026-03-30
Installation doesn't retrieve zero exit code


ascialla:
Installed all the Additional components and then downloaded the repo from git and runned INSTALL.pl script but I runned in an error.

More info in the attached file 

Could someone give me some hints on how to fix the issue? 

Thanks

[vep_fail.txt](https://github.com/user-attachments/files/26350099/vep_fail.txt)

ainefairbrother:
Hi @ascialla, 

Looks like there's a missing dependency. Can you try installing `Test::Warnings`, then re-running `perl INSTALL.pl`?

Best, 

@ainefairbrother

ascialla:
Hi @ainefairbrother  thanks for quick answer. 
Thanks to your suggestion, the test result is now `PASS` . I'm setting the cache now and later I'll try to run VEP example.

Thanks for the help :)

ainefairbrother:
Hi @ascialla, 

Great, thanks for letting me know. 

I'll now close this issue, but please feel free to re-open it or create a new issue if you have any further questions. 

Best, 

@ainefairbrother

ascialla:
I tried to set the cache by following the instructions with VEP installation and I got this message 
```
? 510
 - downloading https://ftp.ensembl.org/pub/release-115/variation/indexed_vep_cache/homo_sapiens_vep_115_GRCh38.tar.gz
 - unpacking homo_sapiens_vep_115_GRCh38.tar.gz
 - converting cache, this may take some time but will allow VEP to look up variants and frequency data much faster
 - use CTRL-C to cancel if you do not wish to convert this cache now (you may run convert_cache.pl later)
2026-03-30 16:37:11 - Processing homo_sapiens
2026-03-30 16:37:11 - Processing version 115_GRCh38
2026-03-30 16:37:11 - No unprocessed types remaining, skipping
2026-03-30 16:37:11 - All done!

```
then I tried to run the demo in examples folder but i got the following error 
`MSG: ERROR: Cache directory /home/admin1/.vep/homo_sapiens not found
`
What am I missing here?

ainefairbrother:
Hi @ascialla, 

Can you show me the VEP command that you ran? 

Thanks, 

@ainefairbrother

ascialla:
I runned the basic command in the tutorial section

~/ensembl-vep$ ./vep -i examples/homo_sapiens_GRCh38.vcf --cache

ainefairbrother:
[...]

PR #1991, opened 2026-03-30
Check for relevant `Test::` modules before running test suite

ainefairbrother:
Related to issue #1990. 

The issue in this case was that, upon install reaching the "Testing VEP installation" part of the script, `Test::Harness` and `Test::Exception` were checked for and found, triggering the full test branch. However, one test dependency,`Test::Warnings`, wasn't checked for, and wasn't available. As such, some of the tests that required it failed, leading to the overall test outcome of `FAIL`. 

This PR adds a check for `Test::Warnings`, and if the 'full test' is skipped due to missing `Test::` modules, a warning is displayed which explicitly lists the missing modules, allowing the user to install them and re-run the setup tests if they would like to. Previously, the 'full test' branch would be silently skipped if any test modules were missing, triggering the 'minimal test' only branch.

### DIFF
--- a/INSTALL.pl
+++ b/INSTALL.pl
@@ -1066,8 +1066,10 @@ sub test() {
 
   print "\nTesting VEP installation\n" unless $QUIET;
 
-  eval q{use Test::Harness; use Test::Exception; };
-  if(!$@) {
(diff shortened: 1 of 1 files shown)


mluypaert (approved):
LGTM

PR #1992, opened 2026-03-30
Fix fork limit off-by-one in convert_cache.pl

rithvik0906:
## Summary
Fixes an off-by-one error in the parallel fork control logic in `convert_cache.pl`.

## Issue
The condition:
  if ($active_forks <= $fork_number)

allowed spawning one extra child process beyond the intended fork limit.

## Fix
Updated condition to:
  if ($active_forks < $fork_number)

This ensures the number of active child processes never exceeds the specified limit.

## Impact
- Prevents resource overuse during parallel execution
- Keeps fork pool size consistent with user-defined `--fork` value
- No API or behavioral changes beyond correcting the limit

## Scope
- Single-line fix in `convert_cache.pl`
- No additional changes included to keep PR minimal and focused

## Testing
Validated with:
perl convert_cache.pl --species hsapiens --version 109 --dir <cache_dir> --fork 4

Confirmed that no extra child processes are spawned beyond the specified fork limit.

### DIFF
--- a/convert_cache.pl
+++ b/convert_cache.pl
@@ -248,7 +248,7 @@ sub main {
           while(keys %$chr_files or $active_forks) {
 
             # only spawn new forks if we have space
-            if($active_forks <= $fork_number) {
+            if($active_forks < $fork_number) {
 
               my $chr = (keys %$chr_files)[0];
               my $files = delete($chr_files->{$chr});


likhitha-surapaneni:
Hi @rithvik0906 , thank you for your time and contribution. Sorry, we could not get back earlier.

This script was originally developed to convert plain-text, chunked variant dumps into a single tabix-indexed file per chromosome. However, we now provide tabix indexed cache files which can be downloaded from the [FTP](https://ftp.ensembl.org/pub/current/variation/indexed_vep_cache/). Hence, we do not plan to support this script going forward, hope this answers your question.

Thanks and regards,
Likhitha

jamie-m-a:
Thanks @rithvik0906 and thank you @likhitha-surapaneni for addressing the PR -  as stated, we no longer support or release unindexed caches, so this script is no longer used. As such I'm closing this PR, but thank you for taking the time to suggest an improvement to our codebase.

PR #1996, opened 2026-04-02
Added option to disable max_sv_size filtering

mluypaert:
[ENSVAR-6982](https://embl.atlassian.net/browse/ENSVAR-6982)

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/Parser.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser.pm
@@ -488,14 +488,16 @@ sub validate_vf {
   }
 
   # check against size upperlimit to avoid memory problems
-  my $len = $vf->{end} - $vf->{start};
-  my $max_sv_size = $self->{max_sv_size};
-  if( $len > $max_sv_size ){
-    $self->skipped_variant_msg(
-      "variant size ($len) is bigger than --max_sv_size ($max_sv_size)"
-    );
-    $vf->{vep_skip} = 1;
-    return 0 if $self->param('rest');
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -488,14 +488,16 @@ sub validate_vf {
...
   # check against size upperlimit to avoid memory problems
-  my $len = $vf->{end} - $vf->{start};
-  my $max_sv_size = $self->{max_sv_size};
-  if( $len > $max_sv_size ){
-    $self->skipped_variant_msg(
-      "variant size ($len) is bigger than --max_sv_size ($max_sv_size)"
-    );
-    $vf->{vep_skip} = 1;
-    return 0 if $self->param('rest');
+  if($self->{max_sv_size} != -1){
```
nakib103, line 491: This is the point I was making in standup. There is no speed benefit if we have to use if we have to check it here for every variant. 

But, I do think there is not a straightforward way of avoiding this check in the loop and not worth time investing.

The `-1` provides easy turn off this feature instead of adding a huge number.

nakib103 (approved):

modules/Bio/EnsEMBL/VEP/Parser.pm
mluypaert: Well, this check saves needing to do the length calculation and one variable assignment for each variant feature, so might make some minor difference. But I agree, the performance difference would probably be barely measurable (if at all).

PR #1997, opened 2026-04-07
[Custom annotation] SV insertion coord fix

nakib103:
https://github.com/Ensembl/ensembl-vep/issues/1645

For insertion the coord should be the flanking bases where the insertion happens. In custom annotation with a VCF file, the end position for an insertion type account for the SVLEN/END and thus add the whole length of the insertion.

Here, we update the end to be re-calculated based on start position in such case.

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/AnnotationSource/File.pm
+++ b/modules/Bio/EnsEMBL/VEP/AnnotationSource/File.pm
@@ -526,7 +526,10 @@ sub _record_overlaps_VF {
   if (defined $self->{_format} && $self->{_format} eq 'vcf' && $parser->get_raw_alternatives =~ /INS/) {
     $ref_end = ($ref_start + length($parser->get_raw_reference) - 1);
   }
-  $ref_start += 1 if defined $self->{_format} && $self->{_format} eq 'bigwig';
+  elsif (defined $self->{_format} && $self->{_format} eq 'bigwig') {
+    # BigWig format is 0-based half-open
+    $ref_start += 1;
+  }
 
   if($type eq 'overlap' || $type eq 'within' || $type eq 'surrounding') {
     # account for insertions in Ensembl world where s = e+1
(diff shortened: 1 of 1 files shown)


ainefairbrother (changes requested):
Hi @nakib103, 

Thanks for this fix. 

It looks like this PR does fix the internal calculation of custom annotation insertion coords and results in correct annotation, but `ins_coords` seems to still include SVLEN. 

My test

Input:
```vcf
##fileformat=VCFv4.2
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
1	100000	same_breakpoint_ins	C	CT	.	PASS	.
1	100020	downstream_ins	A	AT	.	PASS	.
```

Custom:
```vcf
##fileformat=VCFv4.2
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
1	100000	TEST	C	<INS>	.	PASS	SVTYPE=INS;SVLEN=50;END=100050

```

Output (first line only): 
```tsv
#Uploaded_variation	Location	Allele	Gene	Feature	Feature_type	Consequence	cDNA_position	CDS_position	Protein_position	Amino_acids	Codons	Existing_variation	Extra
same_breakpoint_ins	1:100000-100001	T	ENSG00000241860	ENST00000466430	Transcript	intron_variant,non_coding_transcript_variant	-	-	-	-	-	-	IMPACT=MODIFIER;STRAND=-1;ins_coords=1:100001-100050;ins_coords_FILTER=PASS;ins_coords_PC=100.000;ins_id=TEST;ins_id_FILTER=PASS;ins_id_PC=100.000

```

Here, `ins_coords=1:100001-100050` ... should SVLEN be removed from here too?

jamie-m-a:
I tend to agree @ainefairbrother but it's more cosmetic than anything else. I'm taking a look to see if I can amend via a separate PR but I think this one is ok to merge.

ainefairbrother (approved):

PR #1998, opened 2026-04-08
Handle multi-allelic types for SV

likhitha-surapaneni:


modules/Bio/EnsEMBL/VEP/Parser.pm
```diff
@@ -665,6 +663,22 @@ sub get_SO_term {
...
+    my @normalized = map {
+      my $v = $_;
+      $v =~ s/^\s+|\s+$//g;
+      $v =~ s/^\<|\>$//g;
+      uc($v || '');
+    } @parts;
+
+    if (@normalized && scalar(grep { $_ eq $normalized[0] } @normalized) == @normalized) {
+      $type = $parts[0];
+      $type =~ s/^\s+|\s+$//g;
```
mluypaert: Why the whitespace removal after checking if all parts are identical? Is there a possibility of having a non-normalised type string like `< INS>,<INS>` or similar? If so then this whitespace removal should probably happen as part of the `@normalized` array generation to prevent an incorrect assessment of parts differing only on whitespacing. If not then this cleanup is unnessary and should be removed.

mluypaert (changes requested):

mluypaert (commented):
The SV validation at https://github.com/Ensembl/ensembl-vep/blob/9c7c433b13d598647f1013765c504ef3832efdcd/modules/Bio/EnsEMBL/VEP/Parser.pm#L816-L818 should probably be updated to return 0 if `get_SO_term` fails to return a valid term (in other words is not covered by the current mappings)? That way multi-allelic type combinations that are currently not covered by get_SO_term would no longer get outputted with an empty CSQ as was the case with things like DEL,DUP prior to this PR.

likhitha-surapaneni:
> The SV validation at
> 
> https://github.com/Ensembl/ensembl-vep/blob/9c7c433b13d598647f1013765c504ef3832efdcd/modules/Bio/EnsEMBL/VEP/Parser.pm#L816-L818
> 
> should probably be updated to return 0 if `get_SO_term` fails to return a valid term (in other words is not covered by the current mappings)? That way multi-allelic type combinations that are currently not covered by get_SO_term would no longer get outputted with an empty CSQ as was the case with things like DEL,DUP prior to this PR.

Thanks @mluypaert , do you have an example to test this?

mluypaert:
> > The SV validation at
> > https://github.com/Ensembl/ensembl-vep/blob/9c7c433b13d598647f1013765c504ef3832efdcd/modules/Bio/EnsEMBL/VEP/Parser.pm#L816-L818
> > 
> > should probably be updated to return 0 if `get_SO_term` fails to return a valid term (in other words is not covered by the current mappings)? That way multi-allelic type combinations that are currently not covered by get_SO_term would no longer get outputted with an empty CSQ as was the case with things like DEL,DUP prior to this PR.
> 
> Thanks @mluypaert , do you have an example to test this?

The current code covers all the prior failing cases that I'm aware of currently, but if you comment the following two (new) lines
https://github.com/Ensembl/ensembl-vep/blob/9c7c433b13d598647f1013765c504ef3832efdcd/modules/Bio/EnsEMBL/VEP/Parser.pm#L694-L695

Then this case would generate result with an empty CSQ I believe (TBC through testing though):
```
1       149817446       nsv4056484      G       <DEL>,<DUP>     .       .      SOURCE=nstd166;SVLEN=-6000,6000;END=149823445
```

likhitha-surapaneni:
Hi @mluypaert , with the latest fix to skipping lines, this appears to be fixed in my testing due to https://github.com/Ensembl/ensembl-vep/blob/2beada0d57ca6234f467b14a6c60280f4a082717/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm#L479 but please let me know if it is not

mluypaert:
> Hi @mluypaert , with the latest fix to skipping lines, this appears to be fixed in my testing due to
> 
> https://github.com/Ensembl/ensembl-vep/blob/2beada0d57ca6234f467b14a6c60280f4a082717/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm#L479
> but please let me know if it is not

Hi @likhitha-surapaneni Unfortunately it fails exactly as I expected in my testing. I'm using
 * (a copy of) your branch in ensembl-vep with the following local change for testing purpose:
 ```diff
--- a/modules/Bio/EnsEMBL/VEP/Parser.pm
+++ b/modules/Bio/EnsEMBL/VEP/Parser.pm
@@ -690,8 +690,8 @@ sub get_SO_term {
       $subtype = $element if grep /^$element$/i, @mobile_elements;
     }
     $abbrev .= '_' . $subtype;
-  } elsif ($type =~ /DEL/i && $type =~ /DUP/i) {
-    $abbrev = "CNV";
+#    } elsif ($type =~ /DEL/i && $type =~ /DUP/i) {
+#      $abbrev = "CNV";
   } elsif ($type =~ /DUP:TANDEM/i) {
     $abbrev = "TDUP";
   } elsif ($type =~ /CNV:TR/i) {
 ```
 * the latest main in ensembl-variation
 * The following input VCF record (with some additional header lines):
```
1       149817446       nsv4056484      G       <DEL>,<DUP>     .       .      SOURCE=nstd166;SVLEN=-6000,6000;END=149823445
```
 * This VEP config file:
 ```
force_overwrite 1
species homo_sapiens
assembly GRCh38
fasta /nfs/production/flicek/ensembl/variation/luypaert/ENS_beta_hypsipyle_SV/outputs/tmp/Homo_sapiens.GRCh38.dna.toplevel.fa.gz
spdi 1
regulatory 1
pubmed 1
var_synonyms 1
variant_class 1
protein 1
transcript_version 1
allele_number 1
gencode_primary 1
cache_version 114
cache /nfs/production/flicek/ensembl/variation/data/VEP/tabixconverted
offline 1
dir_plugins /hps/software/users/ensembl/variation/luypaert/gitrepos//VEP_plugins
plugin CADD,/nfs/production/flicek/ensembl/variation/enseweb-data_tools/grch38/e114/vep/plugin_data/CADD_prescored_variants.tsv.gz
```



For me that results in the following warning printed to STDERR:
```
WARNING: line 1 skipped (1 149817446 nsv4056484 G <DEL>,<DUP> . . SOURC...): DEL/DUP is not a supported structural variant type
```

and the following output VCF:
```
1       149817446       nsv4056484      G       <DEL>,<DUP>     .       .       SOURCE=nstd166;SVLEN=-6000,6000;END=149823445
```
The output VCF is missing the CSQ tag while having the skipped record in the output (as I expected).

**Review by mluypaert — CHANGES_REQUESTED** [review]

**Comment by likhitha-surapaneni** [comment]
Thanks @mluypaert. This should now be fixed for VCF parsing as there is a logic present to skip lines when SO term is not defined

File: modules/Bio/EnsEMBL/VEP/Parser.pm [hunk 2]
```diff
@@ -690,8 +690,8 @@ sub get_SO_term {
       $subtype = $element if grep /^$element$/i, @mobile_elements;
     }
     $abbrev .= '_' . $subtype;
-  } elsif ($type =~ /DEL/i && $type =~ /DUP/i) {
-    $abbrev = "CNV";
+  # } elsif ($type =~ /DEL/i && $type =~ /DUP/i) {
+  #   $abbrev = "CNV";
```
mluypaert: This needs cleanup (uncommenting), as it was only supposed to be commented for testing purposes, not in the committed code.

modules/Bio/EnsEMBL/VEP/Parser.pm
mluypaert, line 817: As validate_svf is now implemented, this docstring needs updating.

mluypaert: Additionally, this test needs updating as well: https://github.com/Ensembl/ensembl-vep/blob/21d9f098aa5b2bcdcde29cc77703e9a61eb507b7/t/Parser.t#L119

mluypaert (changes requested):

modules/Bio/EnsEMBL/VEP/Parser.pm
likhitha-surapaneni: Thanks @mluypaert , good spot. I added tests and implemented docstrings. Please let me know if it looks good

mluypaert (approved):
LGTM, thanks!

PR #1963, opened 2026-01-12
Fix JSON output for feature end coordinates

likhitha-surapaneni:
[ENSVAR-6961](https://embl.atlassian.net/browse/ENSVAR-6961)
This PR attempts to fix cdna, cds and protein end coordinates where not defined. This can impact REST

### Testing

```
17:g.7668407_7668436del
17:g.7669572_7669647del
17:g.7668407_7669647del
17:g.7668407_7687502del
```

PR #1999, opened 2026-04-09
Fix feature coords for undefined coordinates 

likhitha-surapaneni:
https://github.com/Ensembl/ensembl-vep/pull/1963 attempted to fix coordinates when end is undefined but SNPs may not have end and borrow the end coordinates from start. 
Fixed tests and added a new test
```
17:g.7668407_7668436del
17:g.7669572_7669647del
17:g.7668407_7669647del
17:g.7668407_7687502del

GRCh37
6:g.1102327G>T 
 ```


jamie-m-a (approved):
New regex working as intended, tests amended appropriately - LGTM

Issue #2017, opened 2026-06-12
Two BND records cause incorrect HGVS annotation of an unrelated small variant on the same chromosome


joelarkman:
**Description**

When two inter-chromosomal BND records are present in a VCF, VEP produces incorrect annotation for an unrelated small variant located ~99 Mb away on the same chromosome.

The issue is reproducible both in offline VEP and in the current VEP web interface.

**Minimal reproducing VCF**
```
##fileformat=VCFv4.2 
#CHROM POS ID REF ALT QUAL FILTER INFO 
chr9 37015062 MantaBND:2936:1:2:0:1:0:0 C [chr6:116311629[C . MaxDepth SVTYPE=BND;MATEID=MantaBND:2936:1:2:0:1:0:1;CIPOS=0,1;HOMLEN=1;HOMSEQ=C 
chr9 37020792 MantaBND:4862:0:1:0:3:0:0 C [chr6:136912382[C . MaxDepth SVTYPE=BND;MATEID=MantaBND:4862:0:1:0:3:0:1;CIPOS=0,1;HOMLEN=1;HOMSEQ=C 
chr9 136496196 . CAG C . PASS
```
**Expected result**

The deletion at chr9:136496196 is annotated as:

> NM_017617.5:c.7541_7542del
> p.(Pro2514Argfs*4)

**Observed result**

With both BND records present, VEP produces an incorrect annotation for the deletion.

Examples observed:

_Offline VEP_

> NM_017617.5:c.7543_7544del
> p.(Glu2515ValfsTer3)

_Current VEP web interface_

The deletion is not annotated against NOTCH1 and instead appears as:

> NM_003980.6:c.67+54146_67+54147del
> HGVS_OFFSET=-2

Removing either BND restores the expected annotation.

ainefairbrother:
Hi @joelarkman, 

Thank you for reporting this. I have been able to replicate this issue. We are investigating this and will get back to you soon. 

Many thanks, 

@ainefairbrother

ainefairbrother:
Hi @joelarkman, 

We've now merged a fix for this into the `release/116` branch. 

I will now close this issue. If you have any further issues relating to this, please feel free to re-open the issue or create a new one. 

Many thanks, 

@ainefairbrother

PR #2022, opened 2026-06-29
Fix for BNDs causing mis-annotation of other unrelated variants

ainefairbrother:
This PR aims to fix the bug reported in issue #2017. This bug caused incorrect transcript annotation of unrelated variants in the presence of BND records.

During transcript annotation, VEP identifies input variants that overlap each transcript. This was based only on start/end coords, not chr. BND records make this difficult because they refer to two genomic locs - that of the record and it's mate which is on another chr. VEP includes both when deciding which input variants may overlap a transcript. As such, a BND mate on one chr could cause VEP to consider variants from a different chr if their numeric coords happened to be in the same range. An unrelated variant in the same input file as a BND can thus be incorrectly annotated. 

Example highlighted by #2017: 
  - two BND records on chr9, with mates on chr6
  - a small deletion on chr9

```vcf
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
chr9	37015062	MantaBND:2936:1:2:0:1:0:0	C	[chr6:116311629[C	.	MaxDepth	SVTYPE=BND;MATEID=MantaBND:2936:1:2:0:1:0:1;CIPOS=0,1;HOMLEN=1;HOMSEQ=C
chr9	37020792	MantaBND:4862:0:1:0:3:0:0	C	[chr6:136912382[C	.	MaxDepth	SVTYPE=BND;MATEID=MantaBND:4862:0:1:0:3:0:1;CIPOS=0,1;HOMLEN=1;HOMSEQ=C
chr9	136496196	.	CAG	C	.	PASS	.
```

This resulted in the small deletion being annotated incorrectly, including the wrong hgvs string and the wrong gene, NOTCH1, which is on chr6. 

Changes this PR introduces to fix this behaviour: 

- `InputBuffer::get_overlapping_vfs` 
  - chr additionally supplied as an optional arg
- `Transcript.pm`: 
  - `get_overlapping_vfs` can now optionally take a chromosome, but old usage still works and if no chromosome is passed, behaviour stays the same
   - if a chromosome is passed, returned variants must match that chromosome - this chromosome check is applied to normal variant coordinates, unshifted/original variant coordinates and to BND mate coordinates
  - to handle chr synonyms, chr matching reuses VEP’s existing get_source_chr_name synonym handling sub

Overall, these changes prevent variants on different chromosomes being matched just because their numeric coordinates overlap

### DIFF
--- a/modules/Bio/EnsEMBL/VEP/InputBuffer.pm
+++ b/modules/Bio/EnsEMBL/VEP/InputBuffer.pm
@@ -289,7 +289,6 @@ sub get_overlapping_vfs {
 
   if(@_ == 3) {
     ($chr, $start, $end) = @_;
-    $chr = $self->_normalise_chr_for_vf_overlap($chr);
   }
   else {
     ($start, $end) = @_;
@@ -321,15 +320,14 @@ sub get_overlapping_vfs {
   foreach my $vf (@all_vfs)  {
     my $vf_chr = $vf->{chr};
     $vf_chr ||= $vf->{slice}->{seq_region_name} if $vf->{slice};
-    $vf_chr = $self->_normalise_chr_for_vf_overlap($vf_chr);
 
     if (overlap($vf->{start}, $vf->{end}, $start, $end)){
-      push(@vfs, $vf) if !defined $chr || (defined $vf_chr && $vf_chr eq $chr);
+      push(@vfs, $vf) if $self->_chr_matches_for_vf_overlap($chr, $vf_chr);
     }
 
     if ((defined($vf->{unshifted_end}) && (defined($vf->{unshifted_start})))) {
       if (overlap($vf->{unshifted_start}, $vf->{unshifted_end}, $start, $end)) {
-        push(@vfs, $vf) if !defined $chr || (defined $vf_chr && $vf_chr eq $chr);
+        push(@vfs, $vf) if $self->_chr_matches_for_vf_overlap($chr, $vf_chr);
       }
     }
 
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/VEP/InputBuffer.pm
```diff
@@ -310,26 +318,44 @@ sub get_overlapping_vfs {
...
   }
   return [@vfs];
 }
 
 
+sub _chr_matches_for_vf_overlap {
+  my ($self, $query_chr, $vf_chr) = @_;
+
+  return 1 unless defined $query_chr;
+  return 0 unless defined $vf_chr;
```
nakib103, line 352: @ainefairbrother , is there any reason we are sending 0 when vf chrom is not defined while the other way around for transcript?

ainefairbrother: Hi @nakib103, 

Thanks for reviewing.

`return 1 unless defined $query_chr;` is saying if chr filtering wasn't requested, pass the chr check and add this variant to the results (preserves old behaviour). This is because `get_overlapping_vfs` is used by other parts of vep that call it in the non-chr way.

`return 0 unless defined $vf_chr;` means if chr filtering is being requested, but the variant doesn't have a chr, don't add it to results because we can't confirm the chr - covers an edge case where the candidate variant doesn't have a chr value available.

nakib103: make sense now 👍

nakib103 (approved):

nakib103:
merged to `main`

Issue #2026, opened 2026-07-14
--custom bigWig could use zooms for aggregate stats


davmlaw:
Thanks for the custom summary stats (introduced in #1470)

   For a  `--custom …,format=bigwig,summary_stats=max overlap`, the summary is computed by iterating every stored interval across the variant's reference span:

```
  - Bio::EnsEMBL::IO::Parser::BigWig::seek (Bio/EnsEMBL/IO/Parser/BigWig.pm:63) calls bigWigIntervalQuery($seq, $start, $end) and pushes every returned interval into an in-memory array.
  - Bio::EnsEMBL::VEP::AnnotationSource::File::annotate_InputBuffer (File.pm:316-354) then loops all of them, updating min/max/sum (File.pm:429).
```

  For per-base tracks (phyloP/phastCons) this is ~1 record/base, so cost and memory scale with the interval length. It's invisible on SNVs/small indels but severe on structural variants: on a batch of large SVs (0.5–4.5 Mb) with 4 conservation bigWigs, VEP did not finish in >290 s, versus ~10 s computing the identical max values via the bigWig zoom summaries.

Suggested fix: for large spans, read the interior from the file's precomputed zoom reductions instead of every base. The primitives are already available in the shipped `Bio::DB::BigFile` binding -  `bigWigSummaryArray` /
  `bigWigSummaryArrayExtended` (used in `Bio/EnsEMBL/IO/Adaptor/BigWigAdaptor.pm:109,131`). A naive single-bin zoom query overshoots max where a zoom bin straddles the query boundary, so an exact/zoom split works well: exact  per-base (`bigWigIntervalQuery`) on two narrow end windows of width E, zoom summaries for the interior on a fine grid of bin width G with E > G, so any straddling interior bin is capped inside an exact-scanned end. This is  bit-exact vs the per-base walk in our testing, and also removes the per-base array (lower memory). A small span threshold keeps the current exact path for SNVs/small indels. mean/sum would use the coverage-weighted sumData/validCount from …Extended.

nakib103:
Hi @davmlaw,

Thanks for reporting and the investigation. We will check and get back to you.

Regards,
Nakib

davmlaw:
Hi, I was looking at making a PR and while doing so found #2027 - so might be good idea to fix that first, so we have a truth set to compare against

nakib103:
Make sense, looking forward the fix. Can you make the PR against the `release/116` branch. Thanks!

PR #2029, opened 2026-07-17
Speed up --custom bigWig summary_stats min/max on large spans via zoom summaries

davmlaw:
# Speed up `--custom` bigWig `summary_stats` min/max on large spans via zoom summaries

Relates to Ensembl/ensembl-vep#2026

## Problem

`--custom …,format=bigwig,summary_stats=min` / `max` computes the value by walking
**every stored interval** across the variant's reference span:

- `Bio/EnsEMBL/IO/Parser/BigWig.pm::seek()` → `bigWigIntervalQuery($seq, $start, $end)`
  pushes every interval into an in-memory array.
- `modules/Bio/EnsEMBL/VEP/AnnotationSource/File.pm::annotate_InputBuffer()`
  loops over all of them updating the running min/max.

For per-base tracks (phyloP/phastCons) this is O(span) in time and memory. It is
fine for SNVs and indels, but multi-Mb structural variants effectively never
finish — observed **>290 s vs ~10 s** for an equivalent zoom summary.

## Fix

`Bio::EnsEMBL::VEP::AnnotationSource::File::BigWig` now overrides
`annotate_InputBuffer`. When the requested `summary_stats` are a subset of
`min`/`max`:

- **Small spans** (≤ `zoom_min_span`, i.e. SNVs, indels, small SVs) and every
  non-summary case fall through to the base class unchanged.
- **Large spans** use an aligned exact/zoom decomposition:
  - the two ragged **ends** (width `exact_end`) are read **exactly** with
    `bigWigIntervalQuery`;
  - the **interior** extrema come from the bigWig's precomputed **zoom
    reductions** via `bigWigSummaryArrayExtended` on a grid of width `zoom_grid`.
  - Safety invariant `exact_end > zoom_grid` caps any zoom bin that straddles the
    interior boundary inside an exact-scanned end, so the reported min/max only
    ever reflect a value inside the span. **min and max are bit-exact** vs the
    per-base walk.

Defaults: `zoom_grid = 16384`, `exact_end = 2·zoom_grid = 32768`,
`zoom_min_span = 2·exact_end = 65536`. These are methods (not constants) so they
can be tuned or overridden by a subclass.

This ports the aligned span decomposition from the validated `span_max_aligned`
reference in
[SACGF/variantgrid#1657](https://github.com/SACGF/variantgrid/commit/85456292237d5ff68ed70f4a0a8c8bcd2594228b)
(`annotation/sv_conservation.py`) to Perl.

## Why only min/max

A zoom bin's `minVal`/`maxVal` are exact extrema over the whole bin, so `min`/`max`
are exact. The other statistics are **not** served from zoom and fall back to the
exact per-base path:

- **`sum` / `count`** come out **inflated** from the zoom summaries: the library
  apportions each zoom bin to the query by coordinate overlap and rounds the
  apportioned base count **up to an integer** per bin (`ceil`, see kent
  `bbiRead.c::normalizeCount`); on a real `hg38.phastCons30way.bw` this
  over-counted by ~25–40 %.
- Anything needing `sum`/`mean`/`count` forces the O(span) walk anyway — and once
  you're walking, `min`/`max` come for free — so `summary_stats=max%mean` etc.
  correctly falls back with no loss.

Non-default matching options (`overlap_cutoff`, `distance`, `reciprocal`,
`same_type`, per-record `PC`, or a non-`overlap` type) also fall back.

## Benchmark

Real `hg38.phastCons30way.bw`, `summary_stats=min%max`, exact per-base walk vs the
zoom path:

| span | exact per-base walk | zoom path | min/max |
|------|--------------------|-----------|---------|
| 300 kb | 3,743 ms | 89 ms | identical |
| 400 kb | 5,046 ms | 93 ms | identical |
| 1 Mb   | 12,639 ms | 150 ms | identical |

≈ **42–84× faster**, `min`/`max` bit-exact, across a range of spans.

## Testing

- `t/AnnotationSource_File_BigWig.t`: all existing tests pass unchanged; adds
  large-span tests that annotate a large-span variant through the zoom path and
  assert `min` and `max` against the base-class per-base walk, that the per-record
  output is preserved (same first record and `...` truncation marker), and that a
  combination also needing `sum`/`mean`/`count` falls back to the exact per-base
  path (identical output).
- New fixture `t/testdata/custom/test_large.bw` (~68 kB, 3 kb of per-base scores
  on chr21, 6 zoom levels). A test-only subclass shrinks the window sizes
  (`zoom_grid = 256`) so this modest fixture genuinely exercises the
  decomposition — the interior is served from the reduction-160 zoom level, capped
  by the `exact_end = 512` ends.
- Validated end-to-end against the real `hg38.phastCons30way.bw` through the `vep`
  CLI, and separately confirmed both `min` and `max` bit-exact across 100 kb–2 Mb
  spans on that track.

### Regenerating the fixture

```
# per-base bedGraph over chr21:1000-4000, deterministic values
perl -e 'for(my $p=1000;$p<4000;$p++){ my $h=($p*2654435761)%4294967296;
  printf "chr21\t%d\t%d\t%.3f\n",$p,$p+1,-14+24*($h/4294967296); }' > test_large.bedGraph
printf "chr21\t46709983\n" > chr21.sizes
bedGraphToBigWig test_large.bedGraph chr21.sizes t/testdata/custom/test_large.bw
```

## Edge cases covered

Insertions (span 0 → exact path), start/end-of-contig clamping, spans past the
contig end, and spans straddling or entirely inside a no-data region (a fully
no-data span yields no annotation, exactly as today).


modules/Bio/EnsEMBL/VEP/AnnotationSource/File/BigWig.pm
```diff
@@ -134,8 +150,255 @@ sub valid_chromosomes {
...
+    }
+    else {
+      $self->_annotate_large_span_vf($vf, $stats);
+    }
+  }
+
+  # run the unchanged base-class loop for the rest by pointing it at just them
+  if (@exact_vfs) {
+    my $all = $buffer->buffer;
+    $buffer->buffer(\@exact_vfs);
```
nakib103: this is ok because we are using fork and there is no leaking of buffer between child process. As long as we restore the buffer as annotation sources are run sequentially other sources would not see the affect.

But we should make sure that we do restore the buffer always even in the case of exception (downstream process may not handle well otherwise)

 
```suggestion
    my $ok = eval {
      $buffer->buffer(\@exact_vfs);
      $self->SUPER::annotate_InputBuffer($buffer);
      1;
    };
```

nakib103 (changes requested):

nakib103:
PR sufficiently add functionality to speed up min/max calculation from bigwig annotation. It is implemented only for default parameters (in `_can_use_zoom_summary`) but should work on other condition as well. But lets accept as it is and enhance later with proper tests if need be.

nakib103:
merged to `main`

PR #2032, opened 2026-08-03
add gc tags

jamie-m-a:
Adds the tags, enables the flag_gencode_primary option to work from gff. Working on whether filtering by transcript tag can be enabled.


nakib103 (changes requested):
@jamie-m-a ,
The `--flag_gencode_primary` works fine and flag transcript that are gencode primary.
But filtering with `--gencode_primary` does not work, because although Cache and database both uses https://github.com/Ensembl/ensembl-vep/blob/0e5381d775b6fec0f6ccde7377609a907b88e390/modules/Bio/EnsEMBL/VEP/AnnotationType/Transcript.pm#L200 but GXF does not.

nakib103 (approved):
The PR intention is to add the flag only. The filtering will be in separate PR.
LGTM!~

nakib103:
merged to `main` and `release/116`

PR #2041, opened 2026-08-14
Fixing sort for large headers

jamie-m-a:
Bug reported from web here: https://genomes-ebi.slack.com/archives/C0JUVJV6W/p1786694507190039 

After some digging I discovered it's the sort command that's breaking the header, we hardcoded a value of head -n 1000 when checking for the header in check_VCF.nf, but the file here has a massive amount of headers, so they were truncated and broke downstream processing.

Updated the command to actually check properly for header lines and separate the file for sorting that way.

In doing so it broke again, as there is an unnecessary secondary sort, using much the same command in run_vep.nf, which operates on the split files - but these have already been sorted prior to splitting. So I removed this.

You can test using input that caused the error here: `/nfs/production/flicek/ensembl/variation/webVEP_debug/tmpyq3ina2u/combCF4_459IL_inclDNA25_CFA2og202122232425.EuraNerv_2CAnotRefOrHet_morfarnotHomReforVar_ALLother_NOTHomVarorHet.vcf`

Enable --sort

VEP version 115 / docker 115.2

Making against 115 as there are more changes in 116 version that make it a pain to incorporate there. Will need a manual docker rebuild so web can pull.

### DIFF
--- a/nextflow/modules/check_VCF.nf
+++ b/nextflow/modules/check_VCF.nf
@@ -68,8 +68,9 @@ process checkVCF {
   sort_cmd = ""
   if( params.sort ) {
     isGzipped = vcf.extension == 'gz'
-    cat_cmd   = isGzipped ? "zcat ${vcf}" : "cat ${vcf}"
-    sort_cmd += "(${cat_cmd} | head -1000 | grep '^#'; ${cat_cmd} | grep -v '^#' | sort -k1,1d -k2,2n) > tmp.vcf; "
+    zip_cmd   = isGzipped ? "zcat ${vcf} | " : ""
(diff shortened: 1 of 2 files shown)


nextflow/modules/check_VCF.nf
```diff
@@ -67,9 +67,10 @@ process checkVCF {
 
   sort_cmd = ""
   if( params.sort ) {
-    isGzipped = vcf.extension == 'gz'
-    cat_cmd   = isGzipped ? "zcat ${vcf}" : "cat ${vcf}"
-    sort_cmd += "(${cat_cmd} | head -1000 | grep '^#'; ${cat_cmd} | grep -v '^#' | sort -k1,1d -k2,2n) > tmp.vcf; "
+    isGzipped = vcf.extension.toLowerCase() in ['gz', 'bgz']
```
veidenberg, line 70: This gzip test is quite fragile, I'd use `gzip -t ${vcf}` instead.

PR #2042, opened 2026-08-20
Feature/reg gff

jamie-m-a:
## Summary

PR to support calculation of regulatory consequences from GFFs produced by the regulation team.

### How
- Adds a AnnotationSource/File/RegFeat.pm to VEP, to carryout similar function to the copies under AnnotationSource/Cache and AnnotationSource/Database.
- Adds downstream wiring to AnnotationSourceAdaptor, Runner and OutputFactory.
- Adds a warning to CacheDir.pm informing that gff source overrides cache if user runs with `--regulatory` and `--regulatory_gff`
- Adds params and flags to Config
- Adds various tests and test data

### Running and testing
Test vcf here: `/nfs/production/flicek/ensembl/variation/jma/testing/reg_feat_gff_test.vcf`
Grab [relevant GFFs here](https://ftp.ensembl.org/pub/current/regulation/homo_sapiens/GRCh38/annotation/)

#### Basic version.
`vep -i input.vcf --regulatory_gff file=Homo_sapiens.GRCh38.regulatory_features.v116.gff3.gz`

#### Adding motifs
`vep -i input.vcf --regulatory_gff file=Homo_sapiens.GRCh38.regulatory_features.v116.gff3.gz,motifs=Homo_sapiens.GRCh38.motif_features.v116.gff3.gz`

#### Adding EMARS
`vep -i input.vcf --regulatory_gff file=Homo_sapiens.GRCh38.regulatory_features.v116.gff3.gz,motifs=Homo_sapiens.GRCh38.motif_features.v116.gff3.gz,emars=Homo_sapiens.GRCh38.EMARs.v116.gff.gz`

#### Extended promotors
Running with basic mode at first you'll see the test variants for this have IDs like EXTPROM_only_1
`vep -i input.vcf --regulatory_gff file=Homo_sapiens.GRCh38.regulatory_features.v116.gff3.gz --extended_promotors`

This was designed to take advantage of the promoter extensions defined in the GFF, e.g.
`1       Ensembl promoter        10936   11436   .       .       .       ID=ENSR1_958;extended_start=9953;extended_end=11436;gene_id=ENSG00000290825;gene_name=DDX11L16;gene_biotype=lncRNA;color=#ff0000`

NB EMARs cover the same area, so if you switch on EMARs as well as extended promoters you'll get 2 regulatory annotations for the EXTPROM_only_X test cases.

### Documentation
Separate docs PR under public-plugins: https://github.com/Ensembl/public-plugins/pull/985


modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
```diff
@@ -0,0 +1,665 @@
...
+=cut
+
+sub _get_regfeats_by_coords {
+  my ($self, $c, $s, $e) = @_;
+
+  my $parser = $self->parser();
+
+  my $source_chr = $self->get_source_chr_name($c, 'regfeat', $self->valid_chromosomes);
+
+  return [] unless $parser->seek($source_chr, $s - 1, $e + 1);
```
nakib103: There is a very subtle boundary condition where extended_promoter will fail. The default `cache_region_size` is 1e6, which is used as bin size when fetching records from a annotation source. So if there is no variant for example in bin 0 (0-1_000_000) then no record from that region is fetched. We do not get to apply the extension if the record is not fetched in the first place.

For example, update the `EXTPROM_only_1 ` start to `1019340->919340` and `ENSR1_9DX4` extended start to `1019050->919050` and it will not get annotated with --extended_promoters. 

Some real cases where it might get affected -
ENSR6_BBPQR - start `6:44,000,095` `6:43,999,796` (extended)
ENSR12_9PG55 - end `12:48,999,865` `12:49,000,616` (extended)

Solution is to bypass cache_region_size and load all features?

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
```diff
@@ -0,0 +1,665 @@
...
+  Returntype : hashref
+  Exceptions : none
+  Caller     : Bio::EnsEMBL::VEP::BaseRunner
+  Status     : Stable
+
+=cut
+
+sub info {
+  my $self = shift;
+  return $self->{info} ||= { regulatory_gff => $self->file };
```
nakib103, line 676: Not very important, but do we want to include the other file names (motif, emar) if available?

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
```diff
@@ -0,0 +1,665 @@
...
+regulatory consequences, as an alternative to the funcgen database
+(Bio::EnsEMBL::VEP::AnnotationSource::Database::RegFeat) or the VEP cache
+(Bio::EnsEMBL::VEP::AnnotationSource::Cache::RegFeat).
+
+Records are converted to Bio::EnsEMBL::Funcgen::RegulatoryFeature objects. That
+class is a hard requirement rather than a stylistic choice: consequence
+assignment gates on the feature's Perl class in
+Bio::EnsEMBL::Variation::BaseVariationFeatureOverlapAllele, which blesses a dummy
+into the OverlapConsequence's feature_class and calls isa() on it. A generic
+Bio::EnsEMBL::Feature yields no regulatory consequences and no error.
```
nakib103:
```suggestion
We must create RegulatoryFeature objects because regulatory consequence 
assignment only works for that class. It is gated via the feature_type 
being RegulatoryFeature (see 
Bio::EnsEMBL::Variation::BaseVariationFeatureOverlapAllele::_get_oc_list).
```

a bit more readable?

nakib103 (changes requested):
@jamie-m-a,

Overall looks good. There is one boundary case and some minor changes.
Also, will it be beneficial -
- making `--regulatory-gff` work without cache as a standalone regulatory annotator (`--regulatory` requires cache but this does not)?
- enabling motifs and EMAR without needing to enable regulatory features using `-file`?

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
jamie-m-a: Leaving this, minor.

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
jamie-m-a: There's an extra parameter switched on when running in extended mode that now captures these type of in another bin situations. (the size is empirical - based off the data inthe file, but should be a decent compromise for virtually any situation).

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
jamie-m-a: Agreed but tweaked slightly.

jamie-m-a:
> @jamie-m-a,
> 
> Overall looks good. There is one boundary case and some minor changes. Also, will it be beneficial -
> 
> * making `--regulatory-gff` work without cache as a standalone regulatory annotator (`--regulatory` requires cache but this does not)?
> * enabling motifs and EMAR without needing to enable regulatory features using `-file`?

So it should work alone, doesn't require cache, just can work with it (you can run with --gff / --fasta)
Added the option to run with just emars (and motif).

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
```diff
@@ -0,0 +1,678 @@
...
+  my $parser = $self->parser();
+
+  my $source_chr = $self->get_source_chr_name($c, 'regfeat', $self->valid_chromosomes);
+
+  # pad the query on both sides so a promoter whose core sits in a
+  # neighbouring bin, but whose --extended_promoters reach crosses into this
+  # one, is not invisible to this bin's seek. See $EXTENDED_PROMOTER_MARGIN.
+  my $margin = $self->{extended_promoters} ? $EXTENDED_PROMOTER_MARGIN : 0;
+
+  return [] unless $parser->seek($source_chr, $s - 1 - $margin, $e + 1 + $margin);
```
nakib103:
```suggestion
  my $start = $s - 1 - $margin < 1 ? 1 : $s - 1 - $margin;
  return [] unless $parser->seek($source_chr, $start, $e + 1 + $margin);
```

the start can get 0 or negative and fail to seek.

nakib103 (changes requested):

modules/Bio/EnsEMBL/VEP/AnnotationSource/File/RegFeat.pm
jamie-m-a: Agreed, updated.

nakib103 (approved):
LGTM!~ thanks Jamie

nakib103:
merged to `main` and `release/116`

Issue #2048, opened 2026-09-02
--check_frequency treats an allele frequency of exactly 0 as missing, dropping the rarest variants under --freq_filter include --freq_gt_lt lt


davmlaw:
## Describe the issue

`--check_frequency` treats an allele frequency of exactly `0` as "not observed" rather than as a frequency of zero. Under `--freq_filter include --freq_gt_lt lt` — "keep only variants rarer than X" — this silently drops the rarest possible variants, which are exactly the ones that filter is meant to keep.

In `modules/Bio/EnsEMBL/VEP/AnnotationSource/Cache/BaseCacheVariation.pm`, `_add_check_freq_data_to_vf()` reads:

```perl
my $f = $freq_data->{$alt} || 'NA';
```

A frequency of `0` is false in Perl, so it collapses to the `NA` sentinel and `check_pass()` is never called. The same `||` truthiness appears in the matched-allele fallback (`if($match_b && $freq_data->{$match_b})`), so a zero is swallowed on that path too, and `_freq_check_freqs` reports `NA` instead of `0`.

Behaviour for a variant whose frequency in the chosen population is genuinely `0`, with `--freq_freq 0.05`:

| | current | expected |
|---|---|---|
| `--freq_gt_lt gt --freq_filter exclude` | kept | kept |
| `--freq_gt_lt lt --freq_filter exclude` | kept | dropped |
| `--freq_gt_lt gt --freq_filter include` | dropped | dropped |
| `--freq_gt_lt lt --freq_filter include` | **dropped** | **kept** |

The default path is unaffected: `--filter_common` only sets `check_frequency => 1` and inherits `freq_gt_lt => 'gt'`, `freq_filter => 'exclude'`, where `0 >= threshold` is false either way. Only the `lt` combinations change.

### How common are genuine zeros?

Common enough to matter, especially for sub-populations. `check_frequency_filter()` validates `--freq_pop` against the cache's own column list, so `--freq_pop gnomADe_EAS`, `--freq_pop AFR` and similar are all accepted — and that is where the zeros are. Counting the 1Mb chr21 slice shipped as test data (`t/testdata/cache/homo_sapiens/84_GRCh38/21/25000001-26000000_var.gz`, 116,193 known variants), grouped by whether every allele's frequency in that column is exactly `0`:

| column | all-zero | non-zero |
|---|---|---|
| `gnomADe` (global) | 37 | 1,742 |
| `gnomADe_MID` | 1,779 | 0 |
[...]

PR #2052, opened 2026-09-14
Fix --check_frequency to treat an allele frequency of exactly 0 as non-null

likhitha-surapaneni:
https://github.com/Ensembl/ensembl-vep/issues/2048


nakib103 (approved):
looks good!~

nakib103:
merged to `main`

PR #2054, opened 2026-09-16
Feature/nf vep resources

jamie-m-a:
Updating the resource and retry strategy to ask for less upfront as most jobs will complete with much less memory. Only really runVEP that requires more memory and the vast majority of short variant jobs will complete with 4GB.

Structural variants may require more memory though. 

The runVEP memory allocation defaults to 4 GB, with 4 retries that simply multiple memory by retry number (so 4, 8, 12, 16), but can be set using `--vep_memory`


nextflow/README.md
```diff
@@ -69,6 +69,8 @@ The following config files are used and can be modified depending on user requir
   --filters STRING          Comma-separated list of filter conditions to pass to filter_vep, such as "AF < 0.01,Feature is ENST00000377918".
                             Read more on how to write filters at https://ensembl.org/info/docs/tools/vep/script/vep_filter.html
                             Default: null (filter_vep is not run)
+  --vep_memory STRING       Set the starting amount of memory the runVEP module uses.
```
likhitha-surapaneni, line 72: The same update may be needed in `workflows/run_vep.nf`

likhitha-surapaneni (commented):
Works as expected @jamie-m-a , thank you. Had a minor comment

